PR #12, opened 2014-12-12
fix clusterName substitution

viktortnk:
Currently `clusterName` value replacement result in string with commas

"test-cluster" -> "t,e,s,t,-,c,l,u,s,t,e,r"


pablete:
nice catch! +1

PR #30, opened 2015-02-23
add api

snazy:
PR based on PR #29 

add HTTP API methods

PR #31, opened 2015-02-23
use property file snitch

snazy:
PR based on PR #30 

use property file snitch

optional PR - but I'd recommend to include it

Although it does not any value to "C* on Mesos" yet, we should use the property file snitch. It's easier to add it in a fresh installation that to add it in a live system (and it does not add any harm). Further, we can use this with Mesos rack attribute.


BenWhitehead:
Please rebase on mesosphere:rewrite

BenWhitehead:
Please rebase on current `mesosphere:rewrite`

snazy:
Please ignore this PR for now (superseded by newer PR)

PR #34, opened 2015-02-24
State tracking refactor/rewrite

BenWhitehead:
All state tracking is now done through a new object CassandraCluster

All state and task tracking is now done in CassandraCluster.
This includes but is not limited to:
* Configured Nodes
* Tasks per node
* Health Check history
* Executor Configuration

Each time an offer is received by the scheduler, the scheduler asks the CassandraCluster to optionally provide an executor configuration and tasks that should be launched in that executor.

Configuration values will be read at process startup and then persisted the first time any config values is read.  This means that the values will not be easily influenced by parameters changing across restarts of the framework.

Zk configuration is now read from environment variables so that it can be easily configured.


BenWhitehead:
@snazy Please review

snazy:
`ExecutorCounter` - not sure whether we need such a thing at all. IMO we could simply inject the slave’e IP address into the executor and task IDs since we do not allow multiple C* nodes on one slave.

BTW - not sure whether we should also include the cluster-name in the executor and task IDs (in case there are multiple C* clusters on a Mesos cluster). Or are executor and task IDs unique per framework?

I’m not convinced of always iterating over a big nodes-list from `PersistedCassandraClusterState.nodes()`. That deserialization and iteration would take place multiple times for every single execution of `CassandraScheduler.evaluateOffer()`. This can easily become a huge effort for both the scheduler and zookeeper as more nodes are added to the C* cluster. I’d prefer these changes:
* read the cluster state only once and cache it
* introduce a per-node state object (one state object per node)
* cache these per-node state objects and use CHMs to access nodes, executors and tasks
* only hold references to the nodes in the cluster-state-object

Do we really need that `ExecutorMetadata` message for the `CassandraExecutor`? It does only return the executorID, which is already known by the scheduler.

Despite the issues above, I’d really like to see this PR committed as is and to fix the issues (especially splitting the monolithic state thing) later. IMO it’s easier to fix some of these with the outstanding PRs, that also add some significant changes to the code - I mean, it doesn’t make sense to me to make a PR ”nice” if we know that the affected lines in code will be changed anyway.

BenWhitehead:
`ExecutorCounter` can probably be refactored away. Right now it's being used generate ids used when generating executors for example: `cassandra.ben.node.0.executor`.

Using the current scheme executor and taskIds are namespaced to the cluster. The scheme used for taskIds (e.x. `cassandra.ben.node.0.executor.server`) is as follows: 

1. `cassandra` the base name used for all cassandra framework based tasks
2. `clusterName` (in this case `ben`) is the value provided to `CASSANDRA_CLUSTER_NAME`
3. `node` denotes that this task is for a cassandra node
4. `counter` an integer (sourced from `ExecutorCounter`) to make sure that all nodes have a unique id
5. `executor` denotes that a custom executor is being used (there is probably a scenario in the future where the framework could run tasks that don't run in the custom executor)
6. `server` denotes that the task is a (Cassandra) server task

This structure is not (and should not) be relied upon to derive this information from a systemic standpoint but it does allow for the information to be conveyed to the UI easily for a human to be able to try and understand what's running.

`StatePersistedObject` contains all of the logic and semantics about read/write access to the mesos State Abstraction.  In this scenario `PersistedCassandraClusterState.nodes()` will only be read from ZooKeeper and deserialized the first time the variable is read. One the variable has been read and deserialized it will be stored in an extra internal field `parsedValue` which will be read from there on. This means that the deserialization cost is only incurred once, and whenever a value is (successfully) written, the provided value will then be used for further reads so that in the case of write we only incur the performance impact of serializing the object. Given that `CassandraCluster.evaluateOffer()` is a primarily read heavy set of operations performance impact should be minimal.

Given the design of `CassandraClusterState`, once the cluster is bootstrapped the write load on this object should be minimal.  `CassandraClusterHealthCheckHistory` on the other hand is intentionally it's own object as it is something that will see much more write traffic. (We will also need to add some types and containing object to track schedule tasks but that can be done in a separate PR).

`ExecutorMetadata` is actually serving a couple purposes. The purpose on the surface that we need it to execute for is to retrieve the IP of the host running the executor. `executorId` was added to the object so that we have a symmetric representation that allows the executor to easily reply with all information. The second purpose that the `ExecutorMetatdata` task serves is that it's providing an executor tracking task. There is a long standing bug ([MESOS-313](https://issues.apache.org/jira/browse/MESOS-313)) in mesos where `executorLost` isn't properly implemented so that a scheduler can rely upon it.  By having this metadata task run with the executor, we are able to interpret a loss of this task as a loss of executor so that we can then keep track of the state change in the cluster and properly react to it.


I think I've addressed all the concerns outlined in your review @snazy. I'm going to merge this and I can fix any outstanding issues you see.

PR #37, opened 2015-02-26
non task health check

snazy:
do not use 'launchTasks' for health-check and other status inquiries


BenWhitehead:
Please rebase on current `mesosphere:rewrite`

PR #41, opened 2015-03-02
Add repair+cleanup

snazy:
Adds repair and cleanup incl state management.

Also includes unit tests for both scheduler and executor.


cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -53,39 +44,62 @@
...
+    private final AtomicReference<AbstractNodeJob> currentJob = new AtomicReference<>();
+
+    public CassandraExecutor(ObjectFactory objectFactory) {
+        this.objectFactory = objectFactory;
+    }
 
     @Override
     public void registered(final ExecutorDriver driver, final ExecutorInfo executorInfo, final FrameworkInfo frameworkInfo, final SlaveInfo slaveInfo) {
-        LOGGER.debug("registered(driver : {}, executorInfo : {}, frameworkInfo : {}, slaveInfo : {})", driver, protoToString(executorInfo), protoToString(frameworkInfo), protoToString(slaveInfo));
+        if (LOGGER.isDebugEnabled())
```
BenWhitehead: Please add curly braces to all if statements.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ObjectFactory.java
```diff
@@ -0,0 +1,160 @@
...
+
+    WrappedProcess launchCassandraNodeTask(Marker taskIdMarker, CassandraFrameworkProtos.CassandraServerRunTask cassandraServerRunTask) throws IOException;
+
+    public interface WrappedProcess {
+        void destroy();
+
+        int exitValue();
+    }
+
+    // production object factory - there's another implementation, that's used for mocked unit tests
```
BenWhitehead: Please move to own file but leave package local.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ObjectFactory.java
```diff
@@ -0,0 +1,160 @@
...
+public interface ObjectFactory {
+
+    JmxConnect newJmxConnect(CassandraFrameworkProtos.JmxConnect jmx);
+
+    WrappedProcess launchCassandraNodeTask(Marker taskIdMarker, CassandraFrameworkProtos.CassandraServerRunTask cassandraServerRunTask) throws IOException;
+
+    public interface WrappedProcess {
+        void destroy();
+
+        int exitValue();
```
BenWhitehead: please define custom runtime exception and declare as a throws on this method. Then in the implementation catch the process exception and re-throw as cause of custom execption.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/jmx/NodeCleanupJob.java
```diff
@@ -13,51 +13,73 @@
...
-    }
+    @Override
+    public void startNextKeyspace() {
+        final String keyspace = super.nextKeyspace();
+        if (keyspace == null)
+            return;
 
-    private void finished() {
+        // cleanup is a blocking operation that may take very long
+        new Thread("Cleanup " + keyspace) {
```
BenWhitehead: please move this to being a runnable that is submitted to an executor service, then track the returned future.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/jmx/NodeRepairJob.java
```diff
@@ -83,18 +67,16 @@ public boolean repairNextKeyspace() {
...
+    protected void finished() {
         try {
-            finishedTimestamp = System.currentTimeMillis();
-            long duration = System.currentTimeMillis() - startTimestamp;
-            LOGGER.info("Repair job finished in {} seconds : {}", duration / 1000L, keyspaceStatus);
+            super.finished();
             jmxConnect.getStorageServiceProxy().removeNotificationListener(this);
             close();
-        } catch (ListenerNotFoundException ignored) {
+        } catch (Exception ignored) {
```
BenWhitehead: Please document why you are swallowing all exceptions rather than enumerating the exceptions that are valid to swallow.

cassandra-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutorTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+        repairJob();
+
+        cleanupJob();
+    }
+
+    private void cleanupJob() {
+        CassandraFrameworkProtos.ClusterJobType jobType = CassandraFrameworkProtos.ClusterJobType.CLEANUP;
+
+        assertNull(executor.getCurrentJob());
```
BenWhitehead: Please replace default junit assertions with assertj for extended functionality and easier readability.

http://joel-costigliola.github.io/assertj/

cassandra-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/MockExecutorDriver.java
```diff
@@ -0,0 +1,123 @@
...
+import com.google.protobuf.InvalidProtocolBufferException;
+import org.apache.mesos.Executor;
+import org.apache.mesos.ExecutorDriver;
+import org.apache.mesos.Protos;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+public class MockExecutorDriver implements ExecutorDriver {
```
BenWhitehead: What's the motivation for the mock implementations rather than using something like mockito?

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
```diff
@@ -81,12 +81,13 @@ private static int _main() throws UnknownHostException {
         final int port0 = Integer.parseInt(portOption.get());
 
         final int       executorCount           = Integer.parseInt(     Env.option("CASSANDRA_NODE_COUNT").or("3"));
+        final int       seedCount               = Integer.parseInt(     Env.option("CASSANDRA_SEED_COUNT").or("2"));
```
BenWhitehead: When would we want to allow a user to change the number of seed nodes to be a value other than 3?

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
```diff
@@ -106,19 +107,23 @@ private static int _main() throws UnknownHostException {
...
             resourceMemoryMegabytes,
             resourceDiskMegabytes,
-            healthCheckIntervalSec
+            healthCheckIntervalSec,
+            bootstrapGraceTimeSec
         );
-        final ExecutorCounter executorCounter = new ExecutorCounter(state, 0L);
-        final PersistedCassandraClusterState persistedCassandraClusterState = new PersistedCassandraClusterState(state);
-        final PersistedCassandraClusterHealthCheckHistory persistedCassandraClusterHealthCheckHistory = new PersistedCassandraClusterHealthCheckHistory(state);
+        configuration.numberOfNodes(executorCount);
```
BenWhitehead: these values shouldn't be overwritten here.  If we need to allow for the values to change we will need to create a rest endpoint that can perform the operation and log that it happened.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -91,46 +94,92 @@
...
+        singleton = this;
         this.clock = clock;
         this.httpServerBaseUrl = httpServerBaseUrl;
-        this.execCounter = execCounter;
-        this.clusterState = clusterState;
-        this.healthCheckHistory = healthCheckHistory;
+        this.execCounter = executorCounter;
+        this.clusterState = new PersistedCassandraClusterState(state);
+        this.healthCheckHistory = new PersistedCassandraClusterHealthCheckHistory(state);
+        this.jobsState = new PersistedCassandraClusterJobs(state);
```
BenWhitehead: Please revert this change. `CassandraCluster` should know nothing about how to instantiate the state objects.  If this is to facilitate testing the stateful objects should be instantated against either `org.apache.mesos.state.InMemoryState` or a mock. Remove `singleton`.

This class also shouldn't be used as a service locator.  If another class needs a common dependency with this class we should clean up the dependency tree so that things can be constructed with all dependencies.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -208,28 +277,47 @@ public void recordHealthCheck(@NotNull final String executorId, @NotNull final C
...
     }
 
     @NotNull
     public List<String> getSeedNodes() {
-        return newArrayList(from(clusterState.executorMetadata()).transform(CassandraFrameworkProtosUtils.executorMetadataToIp()));
+        return newArrayList(from(clusterState.nodes())
+                .filter(new Predicate<CassandraNode>() {
+                    @Override
+                    public boolean apply(CassandraNode v) {
+                        return v.getSeed();
```
BenWhitehead: Give this comparison is a constant, please create either a class with a singleton instance or create a static final field with this implementation.

BenWhitehead:
Added some preliminary review feedback.

Please rebase this branch currently has changes in it that will remove things that have been added or that have already been removed.

`git rebase -i upstream/master`. Without a rebase it will be impossible to merge this without breaking other things.

I'll try and spend more time reviewing the changes around the scheduler and offer evaluation tomorrow.

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
snazy: The usual recommendation is: "2 or 3 seed nodes per DC".
But different people have different policies - so it depends on the user. But I don't expect many people to change the default value of 3.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: done

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ObjectFactory.java
snazy: done

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ObjectFactory.java
snazy: done

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/jmx/NodeCleanupJob.java
snazy: done

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/jmx/NodeRepairJob.java
snazy: superseded (only LNFE is necessary)

cassandra-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutorTest.java
snazy: partly replaced some assertions + cleaned up unit tests

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
snazy: done

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: done

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: done

snazy:
Hm - do you really mean `master` branch?
The current "PR" branch merges fine onto `rewrite` branch.

BenWhitehead:
you're right, `rewrite` not `master`

snazy:
Rebased branch in a few commits

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -358,8 +409,34 @@ private CassandraNodeHealthCheckDetails performHealthCheck(@NotNull final Cassan
         return builder.build();
     }
 
-    private static CassandraNodeInfo buildInfo(JmxConnect jmx) throws UnknownHostException {
-        Nodetool nodetool = new Nodetool(jmx);
+    private void processAlive(ExecutorDriver driver) {
```
BenWhitehead: I find the name of this method confusing. Is it supposed to make sure that the process is alive and prevent code from running that requires the process to be alive?  The way the method it calls looks like an assertion to me.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -24,11 +24,20 @@ message CassandraFrameworkConfiguration {
...
+
+message PortMapping {
+    required string name = 1;
+    required int32 port = 2;
 }
 
 message CassandraClusterState {
     repeated CassandraNode nodes = 1;
     repeated ExecutorMetadata executorMetadata = 2;
+    optional int64 lastServerLaunchTimestamp = 3;
```
BenWhitehead: I don't know if we should be using time to determine when to launch servers rather than events and state.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -74,7 +109,11 @@ message CassandraNode {
     optional CassandraNodeTask metadataTask = 6;
     optional CassandraNodeTask serverTask = 7;
 
-    repeated CassandraNodeHealthCheckDetails healthCheckHistory = 8;
+    repeated HealthCheckDetails healthCheckHistory = 8;
+
+    required bool seed = 9;
```
BenWhitehead: I'd prefer to not flag nodes as seeds but instead have a seeds list in the CassandraCluster object. I'm under the impression that seeds won't change often.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -110,6 +149,8 @@ message CassandraNodeTask {
     repeated int64 ports = 6;
 
     optional TaskDetails taskDetails = 7;
+
+    required int64 submittedTimestamp = 8;
```
BenWhitehead: What does this buy us?

cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/CassandraFrameworkProtosUtils.java
```diff
@@ -145,6 +176,15 @@ public String apply(final ExecutorMetadata input) {
         }
     }
 
+    private static final class CassandraNodeToIp implements Function<CassandraNode, String> {
+        private static final CassandraNodeToIp INSTANCE = new CassandraNodeToIp();
+
+        @Override
+        public String apply(final CassandraNode input) {
```
BenWhitehead: Please add `@NotNull` annotation to `input`

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -93,77 +124,35 @@ public void launchTask(final ExecutorDriver driver, final TaskInfo task) {
...
                             nullSlaveStatusDetails()));
                     driver.sendStatusUpdate(taskStatus(task, TaskState.TASK_FINISHED));
                     break;
-                case CASSANDRA_SERVER_HEALTH_CHECK:
-                    final CassandraServerHealthCheckTask healthCheckTask = taskDetails.getCassandraServerHealthCheckTask();
+                case HEALTH_CHECK:
+                    final HealthCheckTask healthCheckTask = taskDetails.getHealthCheckTask();
                     LOGGER.info(taskIdMarker, "Received healthCheckTask: {}", protoToString(healthCheckTask));
-                    final CassandraNodeHealthCheckDetails healthCheck = performHealthCheck(healthCheckTask);
+                    final HealthCheckDetails healthCheck = performHealthCheck(driver);
```
BenWhitehead: Please move `if (serverTaskStatusOutstanding) {...` from `performHealthCheck` out so that `performHealthCheck` doesn't have a dependency on the driver.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -184,158 +173,220 @@ public void launchTask(final ExecutorDriver driver, final TaskInfo task) {
...
                 )
-                .build();
+                    .build();
             final TaskStatus taskStatus = taskStatus(task, TaskState.TASK_ERROR, details);
             driver.sendStatusUpdate(taskStatus);
         }
         LOGGER.debug(taskIdMarker, "< launchTask(driver : {}, task : {})", driver, protoToString(task));
     }
 
+    private void startJob(ExecutorDriver driver, TaskInfo task, TaskDetails taskDetails) {
```
BenWhitehead: All return points of `startJob` should send some sort of taskStatus. Please change this method to return `TaskStatus` which can then be sent by the driver, rather than having many points where `sendStatusUpdate` can be called.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -184,158 +173,220 @@ public void launchTask(final ExecutorDriver driver, final TaskInfo task) {
...
+        if (process == null) {
+            driver.sendStatusUpdate(taskStatus(task, TaskState.TASK_ERROR, nullSlaveStatusDetails()));
```
BenWhitehead: We probably want to have some detail here so the scheduler can intelligently handle if a job couldn't start because a process wasn't running.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -184,158 +173,220 @@ public void launchTask(final ExecutorDriver driver, final TaskInfo task) {
...
+        } else {
+            driver.sendStatusUpdate(taskStatus(task, TaskState.TASK_ERROR, nullSlaveStatusDetails()));
+        }
+    }
+
+    private void jobStatus(ExecutorDriver driver, TaskInfo task) {
+        if (process == null) {
+            LOGGER.error("No Cassandra process to handle node-job-status");
+            if (task != null)
+                driver.sendStatusUpdate(taskStatus(task, TaskState.TASK_ERROR, nullSlaveStatusDetails()));
```
BenWhitehead: We probably want to have some detail here so the scheduler can intelligently handle if a job couldn't start because a process wasn't running.

BenWhitehead: Also this entry assertion is almost exactly the same as what's done in `startJob`. Are (can) they actually be the same and be factored into a helper method to avoid duplication.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -184,158 +173,220 @@ public void launchTask(final ExecutorDriver driver, final TaskInfo task) {
...
+            return;
+        }
+
+        AbstractNodeJob current = currentJob.get();
+        if (current == null) {
+            LOGGER.error("No current job to handle node-job-status");
+            return;
```
BenWhitehead: If there is no current job running we should probably return a status stating such.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -358,8 +409,34 @@ private CassandraNodeHealthCheckDetails performHealthCheck(@NotNull final Cassan
...
+        if (process == null)
+            return;
+        try {
+            int exitCode = process.exitValue();
+            LOGGER.error("Cassandra process terminated unexpectedly with exit code {}", exitCode);
+
+            final SlaveStatusDetails details = SlaveStatusDetails.newBuilder()
+                    .setStatusDetailsType(SlaveStatusDetails.StatusDetailsType.ERROR_DETAILS)
+                    .setSlaveErrorDetails(
+                            SlaveErrorDetails.newBuilder()
```
BenWhitehead: Please make this status object more robust. The scheduler would probably like to be able to take the exit code into consideration when making decisions.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/jmx/NodeRepairJob.java
```diff
@@ -83,15 +70,13 @@ public boolean repairNextKeyspace() {
             commandToKeyspace.put(commandNo, keyspace);
 
             LOGGER.info("Submitted repair for keyspace {} with cmd#{}", keyspace, commandNo);
-            return true;
+            break;
         }
     }
 
-    private void finished() {
+    protected void finished() {
```
BenWhitehead: The method name `finished` to me sounds like it should return a boolean whether the job is finished or not.  If this is supposed to be the "method that is called when job is done" it should probably be renamed to `finish`.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/jmx/ProdJmxConnect.java
```diff
@@ -0,0 +1,281 @@
...
+
+    public StreamManagerMBean getStreamManagerProxy() {
+        if (streamProxy == null) {
+            streamProxy = newProxy(StreamManagerMBean.OBJECT_NAME, StreamManagerMBean.class);
+        }
+        return streamProxy;
+    }
+
+    public CacheServiceMBean getCacheServiceProxy() {
+        if (cacheService == null)
```
BenWhitehead: Missing `{` `}` on if statement

cassandra-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutorTest.java
```diff
@@ -0,0 +1,281 @@
...
+        assertNotNull(executor.getCurrentJob());
+        // MBean emits system KS + 3 other KS; one KS is currently processing; 2 KS remaining
+        assertEquals(2, executor.getCurrentJob().getRemainingKeyspaces().size());
+        assertFalse(executor.getCurrentJob().isFinished());
+        assertFalse(objectFactory.storageServiceProxy.listeners.isEmpty());
+
+        objectFactory.storageServiceProxy.emitRepairNotification(ActiveRepairService.Status.STARTED);
+        objectFactory.storageServiceProxy.emitRepairNotification(ActiveRepairService.Status.SESSION_SUCCESS);
+        objectFactory.storageServiceProxy.emitRepairNotification(ActiveRepairService.Status.SESSION_SUCCESS);
+        objectFactory.storageServiceProxy.emitRepairNotification(ActiveRepairService.Status.SESSION_SUCCESS);
```
BenWhitehead: What is the significance to calling this three times with the same argument? Please add a comment explaining.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -502,4 +681,167 @@ private static CassandraNodeTask getMetadataTask(@NotNull final String executorI
...
+            configuration.numberOfNodes(nodeCount);
+        } catch (IllegalArgumentException e) {
+            LOGGER.info("Cannout update number-of-nodes", e);
+        }
+        return configuration.numberOfNodes();
+    }
+
+    // cluster tasks
+
+    public void handleClusterTask(String executorId, TasksForOffer tasksForOffer) {
```
BenWhitehead: Make this method private.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -74,7 +109,11 @@ message CassandraNode {
     optional CassandraNodeTask metadataTask = 6;
```
BenWhitehead: Why does `CassandraNode` have a `JmxConnect` I don't think it should be any higher than executor in the hierarchy and really is probably more an attribute for the `CassandraNodeTask`.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/PersistedCassandraClusterState.java
```diff
@@ -78,8 +86,35 @@ public void executorMetadata(@NotNull final List<CassandraFrameworkProtos.Execut
...
+                nodeList.set(i, node);
+                nodes(nodeList);
+                return;
+            }
+        }
+        nodeList.add(node);
+        nodes(nodeList);
+    }
+
+    public Tuple2<Integer, Integer> nodeCounts() {
```
BenWhitehead: Given that both items are Integer it's will be confusing trying to remember which tuple value is seeds and which is nodes. Create a more expressive type to be the return type for this method.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -502,4 +681,167 @@ private static CassandraNodeTask getMetadataTask(@NotNull final String executorI
...
+                    .setJobType(currentTask.getJobType())
+                    .setTaskId(cassandraNodeTask.getTaskId())
+                    .setStartedTimestamp(cassandraNodeTask.getSubmittedTimestamp())
+                    .build();
+            jobsState.nextNode(currentTask, currentNode);
+
+            LOGGER.info("Starting cluster job {} on {}/{}", currentTask.getJobType().name(), nextNode.get().getIp(),
+                    nextNode.get().getHostname());
+
+            return;
```
BenWhitehead: pointless return statement

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -89,6 +106,20 @@
     private final PersistedCassandraClusterHealthCheckHistory healthCheckHistory;
     @NotNull
     private final PersistedCassandraFrameworkConfiguration configuration;
+    @NotNull
+    private final PersistedCassandraClusterJobs jobsState;
+
+    public static int getPortMapping(CassandraFrameworkConfiguration configuration, String name) {
```
BenWhitehead: I'm not crazy about this method it creates odd coupling between this and ApiController.

We can revisit in latter though.

BenWhitehead:
Thanks for the squashed commits! Made reviewing much easier. Several points noted above.

I'm also seeing an issue where the server task isn't launched and I get a message like the following in the log:
```
Cannot launch non-seed node (seed node requirement not fulfilled)
```
I'm seeing this error despite trying to launch my first node.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: Hm - that would duplicate the `if`-logic to two places (`frameworkMessage` + `launchTask`). Or we could agree that we will perform heath checks only via framework messages - I'd be fine with that.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: done

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: Added an error-code field to the protobuf message for that reason (commit 7d080c8)

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: 1st comment: done with 7d080c8
2nd comment: i've factored out the duplicate code. At one point we have to check if there's a task available (framework message). commit: abc8edb

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: Hm - yes. But this is a bigger task. I've added a task to my plate: "cluster/node job status handling, reply from node on error-conditions, handle node job timeouts, etc". IMO we should think about that more carefully as possible in a review.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: Yep. I've changed the method name and added some javadoc. It's a bit difficult (for me?) to find an expressive method name. (commit 79313aa)

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: Added a new error-code enum for this one (commit 86a9f66)

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/jmx/NodeRepairJob.java
snazy: Renamed to `cleanupAfterJobFinished` (commit df69053)

cassandra-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutorTest.java
snazy: ok (commit dd01795)

cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/CassandraFrameworkProtosUtils.java
snazy: yup - also added some more of these (commit fda3b83)

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: No - timestamp is correct - it indicates the timestamp when the server has been launched (so we can to grace/pause time handling)

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: No - I think we should keep it there because we should keep it ”constant” over executor restarts. Gets a bit more important if there's JMX authentication in place ... sometimes.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: Can we move that to a separate PR?

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: TBH - I don't remember. Removed (commit 8d135a2)

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: ?? Do you mean `getPortMapping` or the `ApiController` c'tor? The c'tor is really ugly - but less uglier than a singleton in `CassandraCluster`.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: done (commit 88e415b)

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: fixed

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/PersistedCassandraClusterState.java
snazy: done (commit 1cd424b)

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/jmx/ProdJmxConnect.java
snazy: done (commit fa8e75d)

snazy:
I'll explain the `Cannot launch non-seed node (seed node requirement not fulfilled)` a bit more detailed:
Initially the logic requires that as many `CassandraNode` have been built as seed nodes are requested. It does not require a metadata or server task to exist.
It just makes sure that we have a IP addresses of all seed nodes. That's required to build the `cassandra.yaml`file.
For a test environment it should be appropriate to set the number of seed nodes to 1 (instead of leaving the default of 2). Means: if you can only start one node, it will never start anything - because the IP addresses of these 2 seed nodes can never be inquired.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
BenWhitehead: sure.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
BenWhitehead: ok

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
BenWhitehead: I think health check via `frameworkMessage` is probably sufficient

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
BenWhitehead: Ok, lets move it to a future PR

BenWhitehead:
Okay, please add a note into the readme that a developer should set `CASSANDRA_SEED_COUNT` when in development.

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
```diff
@@ -106,19 +108,21 @@ private static int _main() throws UnknownHostException {
             matcher.group(2)
         );
 
+        if (seedCount > executorCount || seedCount <= 0 || executorCount <= 0)
```
BenWhitehead: Missing `{` `}` for if statement

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -206,99 +285,200 @@ public void recordHealthCheck(@NotNull final String executorId, @NotNull final C
...
-//            if (ia.isLoopbackAddress())
-//                try (ServerSocket serverSocket = new ServerSocket(0)) {
-//                    jmxPort = serverSocket.getLocalPort();
-//                } catch (IOException e) {
-//                    throw new RuntimeException(e);
-//                }
+
+            int jmxPort = getPortMapping(PORT_JMX);
+            if (ia.isLoopbackAddress()) {
+                try (ServerSocket serverSocket = new ServerSocket(0)) {
```
BenWhitehead: Instead of picking a port from the host, a port from the Offer should be used (especially because this port is being evaluated on the node running the framework and not necessarily the node running the server)

snazy: The code's only meaningful for testing/developing on "localhost" - don't think it's worth adding much complexity here.

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
snazy: done (commit b6af309)

snazy:
Updated readme (commit 59001f7) along change to recommended bootstrap grace time.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: Alright, HC (and job status) only via framework messages (commit 60499ad)

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -13,18 +13,448 @@
...
+
+    public ApiController(CassandraCluster cluster) {
+        this.cluster = cluster;
+    }
+
+    @GET
+    @Path("/")
+    @Produces("text/html")
+    public String indexPage() {
+        return "<a href=\"status\">All nodes</a> <br/>" +
```
BenWhitehead: Please return a json object here rather than malformed html. It could just be a map.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -13,18 +13,448 @@
...
+        StringWriter sw = new StringWriter();
+        try {
+            CassandraFrameworkProtos.CassandraFrameworkConfiguration configuration = cluster.getConfiguration().get();
+
+            JsonFactory factory = new JsonFactory();
+            JsonGenerator json = factory.createGenerator(sw);
+            json.setPrettyPrinter(new DefaultPrettyPrinter());
+            json.writeStartObject();
+
+            json.writeNumberField("native_port", CassandraCluster.getPortMapping(configuration, CassandraCluster.PORT_NATIVE));
```
BenWhitehead: Please update all json properties to use camel case instead of snake case.  This is nice from the JS side of things in that a UI can then stick with camel case convention when accessing properties.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -13,18 +13,448 @@
...
+
+            json.writeNumberField("next_server_launch_at", cluster.nextPossibleServerLaunchTimestamp());
+
+            json.writeArrayFieldStart("nodes");
+            for (CassandraFrameworkProtos.CassandraNode cassandraNode : clusterState.getNodesList()) {
+                json.writeStartObject();
+
+                writeTask(json, "server_task", cassandraNode.getServerTask());
+                writeTask(json, "metadata_task", cassandraNode.getMetadataTask());
+                json.writeBooleanField("has_node_executor", cassandraNode.hasCassandraNodeExecutor());
```
BenWhitehead: please remove this property `has_node_executor` it's not buying us anything.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -13,18 +13,448 @@
...
+        json.writeArrayFieldStart("seeds");
+        for (String seed : cluster.getSeedNodes())
+            json.writeString(seed);
+        json.writeEndArray();
+    }
+
+    @GET
+    @Path("/status")
+    @Produces("application/json")
+    public Response status() {
```
BenWhitehead: Can you explain the motivation for the object structure that is created here rather than just creating a wrapper object and returning the underlying protobuf objects?

BenWhitehead: Actually I just tried return a protobuf message directly and it errored out with a cycle, so that answers that part of the question.  But why did you change the structure rather than something like:
```json
{
  "config": {
    "frameworkId": "",
    "frameworkName": "",
    "cassandraVersion": "",
    "numberOfNodes": "",
    "cpuCores": "",
    "memMb": "",
    "diskMb": "",
    "healthCheckIntervalSeconds": "",
    "numberOfSeeds": "",
    "bootstrapGraceTimeSeconds": "",
    "portMapping": [...]
  },
  "clusterState": {
    "nodes": [...],
    "executorMetadata": [...],
    "lastServerLaunchTimestamp": 123123123123
  }
}
```

BenWhitehead: Actually now that I think about it more, lets break these into separate resources.

one for config, one for cluster nodes and one for health check info.

I don't think we should embed health check info in the main object

BenWhitehead:
Added a couple points after playing around with the API, perhaps the object structure stuff can be deferred to another PR, but the points about return json from `/` and changing property names to camelcase need to be done before merge.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
snazy: Hm - that's intended for humans to get an idea what's available.

snazy:
OK, committed 18f511a with the camel case change. Regarding `/` I'd like to leave it as is - it's intended for humans to get a quick understanding about the API.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -13,18 +13,494 @@
...
+
+    private final CassandraCluster cluster;
+
+    public ApiController(CassandraCluster cluster) {
+        this.cluster = cluster;
+    }
+
+    @GET
+    @Path("/")
+    @Produces("text/html")
```
BenWhitehead: Remove `text/html` since content type is set directly in response object.

BenWhitehead:
one small nit, then ready to merge I think.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
snazy: TBH: a stupid oversight :)

BenWhitehead: no worries

BenWhitehead:
Investigating a build environment issue before merging to master as [build](https://teamcity.mesosphere.io/viewLog.html?buildId=12102&tab=buildResultsDiv&buildTypeId=Oss_Mesos_Cassandra_CassandraFramework) is failing right now

BenWhitehead:
Build is passing now, had to exclude coverage info for generated protos `io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos*`

PR #44, opened 2015-03-10
Cassandra memory calculation

snazy:
Memory used by C* is can be roughly categorizes into:
* Java heap
* Off heap (more detailed in readme)
* OS block cache

In the simplest mode (just providing `memMb`), i.e. `javaHeap = offHeap = memMb / 2`. Memory sizing will always work if at least one of `memMb`, `memJavaHeapMb` or `memAssumeOffHeapMb` is provided.
A real production deployment however should include at least `memMb` and `memJavaHeapMb` according to the real workload. Even better: provide descent values for all (required) configuration options (environment + `cassandra.yaml`) and memory sizing.

Factored out per-DC / per-node specific configuration to `CassandraConfigRole`. At the moment it does not have any useful meaning (except moving config to a specialized structure). But we can use it in the future to scale-up nodes (e.g. double cores, double ram) one-after-another or to define different settings by DC. In the future it should would by assigning a "config-role-name" to a DC or node and lookup the config-role in `CassandraFrameworkConfiguration.configRoles`.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -94,6 +94,39 @@ The Cassandra Mesos Framework requires an install of Maven between 3.2.1 and 3.2
 
 _NOTE: Maven 3.2.5 has a binary incompatible change with earlier version of Maven 3.2.x and will cause an exception to be thrown when the protobuf tool chain tries to run._
 
+## Cassandra memory usage
+
+Memory used by Cassandra can be roughly categorized into:
+
+* Java heap memory. The amount of memory used by the Java VM for heap memory.
+* Off heap memory. Off heap is used for several reasons by Cassandra:
+     * **index-summary** (default: 5% of the heap size)
+       configured in `cassandra.yaml` - see `index_summary_capacity_in_mb`
+       default to 5% of the heap size (may exceed)
+     * **key-cache** (default: 5% of the heap size)
+       configured in `cassandra.yaml` - see `key_cache_size_in_mb`
+       default to 5% of the heap size
+     * **row-cache** (default: off)
+       configured in `cassandra.yaml` - see `row_cache_size_in_mb` (must be explicitly enabled in taskEnv)
+       default to 0
+     * **counter-cache** (default: min(2.5% of Heap (in MB), 50MB))
+       configured in `cassandra.yaml` - see `counter_cache_size_in_mb`
+       default: min(2.5% of Heap (in MB), 50MB) ; 0 means no cache
+     * **memtables** (default on-heap)
+       configured in `cassandra.yaml` - see `file_cache_size_in_mb`
+       default to the smaller of 1/4 of heap or 512MB
+     * **file-cache** (default: min(25% of Heap (in MB), 512MB))
+       configured in `cassandra.yaml` - see `file_cache_size_in_mb`
+       default to the smaller of 1/4 of heap or 512MB
+     * overhead during flushes/compactions/cleanup
+       implicitly defined by workload
+* OS buffer cache. The amount of (provisioned) memory reserved for the operating system for disk block buffers.
+
+The default configuration simply assumes that you need as much off-heap memory than Java heap memory.
+It basically divides the provisioned amount of memory by 2 and assigns it to the Java heap.
+
+A good planned production system is sized to meet its workload requirements. That does mean proper values for
+Cassandra process environment, `cassandra.yaml` and memory sizing.
 
 ### Setup maven toolchain for protoc
 
--- a/README.markdown
+++ b/README.markdown
@@ -128,6 +128,12 @@ It basically divides the provisioned amount of memory by 2 and assigns it to the
 A good planned production system is sized to meet its workload requirements. That does mean proper values for
 Cassandra process environment, `cassandra.yaml` and memory sizing.
 
+You should not run Cassandra (even in test environments) with less than 4 GB configured in `memMb`.
+A recommended minimum value for `memMb` is 16GB. In times where RAM is getting cheaper, provision as much as you can
+afford - with 8 to 16 GB for `memJavaHeapMb`. Remember to figure out the really required numbers in load and
+stress tests with your application.
+
+
 ### Setup maven toolchain for protoc
 
 1. Download version 2.5.0 of protobuf [here](https://code.google.com/p/protobuf/downloads/list)
--- a/cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
+++ b/cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
@@ -119,7 +119,6 @@ public final class Main {
             .setNumberOfNodes(executorCount)
             .setNumberOfSeeds(seedCount)
             .setMemMb(resourceMemoryMegabytes);
-        CassandraCluster.fillConfigRoleGaps(defaultConfigRole);
         final PersistedCassandraFrameworkConfiguration configuration = new PersistedCassandraFrameworkConfiguration(
             state,
             frameworkName,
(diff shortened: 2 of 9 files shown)


cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -18,15 +18,75 @@ option java_outer_classname = "CassandraFrameworkProtos";
...
-    optional int64 diskMb = 7;
     optional int64 healthCheckIntervalSeconds = 8;
-    optional int32 numberOfSeeds = 9;
     optional int64 bootstrapGraceTimeSeconds = 10;
     repeated PortMapping portMapping = 11;
+
+    // default configuration
+    required CassandraConfigRole defaultConfigRole = 6;
+    // additional configurations (currently unused)
+    repeated CassandraConfigRole configRoles = 7;
```
BenWhitehead: If this isn't being used lets remove it. Protobuf is tolerant of fields being added later.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -18,15 +18,75 @@ option java_outer_classname = "CassandraFrameworkProtos";
...
+
+    // total memory required (must be greater than memJavaHeapMb + memAssumeOffHeapMb)
+    optional int64 memMb = 5;
+
+    // Cassandra memory usage can be categorized into
+    // 1. Java heap (MAX_HEAP_SIZE) - including new-gen (HEAP_NEWSIZE)
+    //    amount of Java heap in MB
+    //    (defaults to 50% of memMb, if not present)
+    optional int64 memJavaHeapMb = 8;
+    // 2. Off-Heap (roughly 50% of Java heap)
```
BenWhitehead: Inconsistent comment. Should off heap space be 50% of what the java heap is, or 50% of the memory allocation from mesos?

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -18,15 +18,75 @@ option java_outer_classname = "CassandraFrameworkProtos";
...
+    //    amount of Off heap in MB
+    //    (defaults to 50% of memMb, if not present)
+    optional int64 memAssumeOffHeapMb = 9;
```
BenWhitehead: Why is this named `memAssumeOffHeapMb` instead of `memOffHeapMb`?

snazy: It's just an assumption. We cannot exactly constraint off-heap memory usage since some "bits" are allocated on-demand when needed (temporarily).

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: fixed (9dc53a2)

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: removed (f189925)

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
BenWhitehead: But from a resource accounting perspective it's what we're allocating as space that isn't part of the heap but is still within the memory limit of the cgroup mesos creates.

BenWhitehead:
Please rebase or merge changes up. I suspect it's readme changes preventing GH fro being able to merge.

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
```diff
@@ -112,15 +112,18 @@ private static int _main() throws UnknownHostException {
...
         }
 
+        CassandraFrameworkProtos.CassandraConfigRole.Builder defaultConfigRole = CassandraFrameworkProtos.CassandraConfigRole.newBuilder()
+            .setCassandraVersion(cassandraVersion)
+            .setCpuCores(resourceCpuCores)
+            .setDiskMb(resourceDiskMegabytes)
+            .setNumberOfNodes(executorCount)
+            .setNumberOfSeeds(seedCount)
+            .setMemMb(resourceMemoryMegabytes);
+        CassandraCluster.fillConfigRoleGaps(defaultConfigRole);
```
BenWhitehead: Defining the default role should be done in the default `Supplier` in [`PersistedCassandraFrameworkConfiguration`](https://github.com/mesosphere/cassandra-mesos/blob/rewrite/cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/PersistedCassandraFrameworkConfiguration.java#L46) rather than calling it in Main

README.markdown
```diff
@@ -55,6 +55,41 @@ This can be done by passing the following flag to the mesos-slave process:
...
+       default to the smaller of 1/4 of heap or 512MB
+     * overhead during flushes/compactions/cleanup
+       implicitly defined by workload
+* OS buffer cache. The amount of (provisioned) memory reserved for the operating system for disk block buffers.
+
+The default configuration simply assumes that you need as much off-heap memory than Java heap memory.
+It basically divides the provisioned amount of memory by 2 and assigns it to the Java heap.
+
+A good planned production system is sized to meet its workload requirements. That does mean proper values for
+Cassandra process environment, `cassandra.yaml` and memory sizing.
```
BenWhitehead: Great explanation, it might be good to add a recommended minimum when running in production or a link to some cassandra documentation that talks about deciding.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -818,4 +829,28 @@ public ClusterJobStatus getLastClusterJob(ClusterJobType jobType) {
         }
         return null;
     }
+
+    public static void fillConfigRoleGaps(CassandraConfigRole.Builder configRole) {
```
BenWhitehead: please move into `PersistedCassandraFrameworkConfiguration` when updating the default supplier.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/PersistedCassandraFrameworkConfiguration.java
```diff
@@ -45,12 +40,7 @@ public PersistedCassandraFrameworkConfiguration(
                 public CassandraFrameworkConfiguration get() {
                     return CassandraFrameworkConfiguration.newBuilder()
                         .setFrameworkName(frameworkName)
-                        .setCassandraVersion(cassandraVersion)
-                        .setNumberOfNodes(numberOfNodes)
-                        .setNumberOfSeeds(numberOfSeeds)
-                        .setCpuCores(cpuCores)
-                        .setMemMb(memMb)
-                        .setDiskMb(diskMb)
+                        .setDefaultConfigRole(defaultConfigRole)
```
BenWhitehead: Move building of default config role to here.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: Right. The default off-heap requirement is about 50% of the Java heap. So using 50% of the provisioned `memMb` should work. That's still too few IMO since we really want to have room left for block cache.

snazy:
Alright, fixed the latest issues and rebased the branch.

PR #45, opened 2015-03-10
C node restart

snazy:
Adds:
* Cassandra server shutdown
* Cassandra server start
* Cassandra server restart

To do that, a `TargetRunState` enum is introduced. It knows three states: `RUN`=node has to run, `STOP`=node should not run, `RESTART`=stop and start node (sets state to `RUN` afterwards)

Since a new task ID (`shutdown`) is being introduced (and a follow-up PR introduces a new state `TERMINATE` along with another new task `terminate`) I've change refactored the `metadataTask` and `serverTask` fields to a general-purpose list of `CassandraNode.tasks`.

### DIFF
--- a/cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
+++ b/cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
@@ -441,6 +441,9 @@ public final class CassandraExecutor implements Executor {
 
             forceCurrentJobAbort();
 
+            // send health check that process is no longer alive
+            healthCheck(driver);
+
         } catch (IllegalThreadStateException e) {
             // ignore
         }
--- a/cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
+++ b/cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
@@ -30,6 +30,7 @@ import org.slf4j.MarkerFactory;
 
 import java.net.UnknownHostException;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
 import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.*;
@@ -52,7 +53,6 @@ public final class CassandraExecutor implements Executor {
     // But without I/O redirection. Note: we need an Oracle JVM then - not a big deal - C* requires an Oracle JVM.
     private volatile WrappedProcess process;
     private volatile JmxConnect jmxConnect;
-    private volatile boolean serverTaskStatusOutstanding;
 
     private ExecutorInfo executorInfo;
     private TaskInfo serverTask;
@@ -60,6 +60,8 @@ public final class CassandraExecutor implements Executor {
     private final AtomicReference<AbstractNodeJob> currentJob = new AtomicReference<>();
 
     private final ExecutorService executorService;
+
+    private final AtomicBoolean killDaemonSingleton = new AtomicBoolean();
 
     public CassandraExecutor(ObjectFactory objectFactory) {
         this.objectFactory = objectFactory;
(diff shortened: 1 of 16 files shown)


snazy:
Note: the "replace node PR" depends on this PR - so I'll wait until this one's been merged.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -124,19 +124,51 @@ public void launchTask(final ExecutorDriver driver, final TaskInfo task) {
...
+                                        LOGGER.info(taskIdMarker, "Cassandra process terminated with exit code {}", exitCode);
+                                        break;
+                                    } catch (IllegalThreadStateException e) {
+                                        // ignore
+                                    }
+                                }
+                            } catch (Exception e) {
+                                LOGGER.error(taskIdMarker, "Failed to stop Cassandra daemon - forcing process destroy", e);
+                            }
+                        }
```
BenWhitehead: we probably want to add an `else` here that will fall back to trying to run `kill -9 $(cat cassandra.pid)` if we're not able to use the jmx `stopDaemon` method.

We should also remove `cassandra.pid` when we shutdown the server.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -343,6 +379,9 @@ private void safeShutdown() {
                 // ignore this
             }
             process = null;
+
+            // send health check that process is no longer alive
+            healthCheck(driver);
```
BenWhitehead: What is the motivation for sending a healthcheck message when we're shutting down the server.  Isn't "shut down" and "unhealthy" two different things in the scenario that the scheduler asked that the process be shut down?

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -489,4 +496,62 @@ private void writeClusterJobNode(JsonGenerator json, CassandraFrameworkProtos.No
 
         json.writeEndObject();
     }
+
+    // node run / stop / restart
+
+    @GET
```
BenWhitehead: Change to `@POST` all write operations should be `@POST` or `@PUT`

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -489,4 +496,62 @@ private void writeClusterJobNode(JsonGenerator json, CassandraFrameworkProtos.No
...
+    @Path("/node/stop/{node}")
+    public Response nodeStop(@PathParam("node") String node) {
+        CassandraFrameworkProtos.CassandraNode cassandraNode = cluster.nodeStop(node);
+
+        return nodeStatusUpdate(cassandraNode);
+    }
+
+    @GET
```
BenWhitehead: Change to `@POST` all write operations should be `@POST` or `@PUT`

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -489,4 +496,62 @@ private void writeClusterJobNode(JsonGenerator json, CassandraFrameworkProtos.No
...
+    @Path("/node/run/{node}")
+    public Response nodeStart(@PathParam("node") String node) {
+        CassandraFrameworkProtos.CassandraNode cassandraNode = cluster.nodeRun(node);
+
+        return nodeStatusUpdate(cassandraNode);
+    }
+
+    @GET
```
BenWhitehead: Change to `@POST` all write operations should be `@POST` or `@PUT`

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -158,24 +158,27 @@ public void removeTask(@NotNull final String taskId, Protos.TaskStatus status) {
...
+            CassandraNodeTask nodeTask = CassandraFrameworkProtosUtils.getTaskForNode(cassandraNode, taskId);
+            if (nodeTask == null) {
                 newNodes.add(cassandraNode);
+                continue;
+            }
+            CassandraNode.Builder builder = CassandraFrameworkProtosUtils.removeTask(cassandraNode, nodeTask);
+            changed = true;
+            switch (nodeTask.getTaskType()) {
+                case METADATA:
+                    // TODO shouldn't we also assume that the server task is no longer running ??
```
BenWhitehead: Removing of executor info and taking the server with it should be handled by the status update from `executorLost` so that once `executorLost` is fixed in mesos it can still be handled there.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -158,24 +158,27 @@ public void removeTask(@NotNull final String taskId, Protos.TaskStatus status) {
...
+                    // TODO do we need to remove the executor metadata ??
+
+                    removeExecutorMetadata(nodeTask.getExecutorId());
+                    builder.clearTasks();
+                    break;
+                case SERVER:
```
BenWhitehead: redundant switch branch, `changed` is set to `true` before the switch begins.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -270,7 +272,7 @@ public void recordHealthCheck(@NotNull final String executorId, @NotNull final H
             if (nodeOpt.isPresent()) {
                 LOGGER.info(
                     "health check result unhealthy for node: {}. Message: '{}'",
-                    nodeOpt.get().getCassandraNodeExecutor().getExecutorId(),
+                    nodeOpt.get().hasCassandraNodeExecutor() ? nodeOpt.get().getCassandraNodeExecutor().getExecutorId() : "??",
```
BenWhitehead: A better fallback would be `nodeOpt.get().getHostname()` instead of `"??"`

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -330,14 +332,29 @@ public TasksForOffer getTasksForOffer(@NotNull final Protos.Offer offer) {
...
             } else {
                 final Optional<ExecutorMetadata> maybeMetadata = getExecutorMetadata(executorId);
                 if (maybeMetadata.isPresent()) {
-                    if (!node.hasServerTask()) {
+                    if (serverTask == null) {
+                        switch (node.getTargetRunState()) {
+                            case RUN:
+                                break;
+                            case STOP:
+                                if (LOGGER.isDebugEnabled())
```
BenWhitehead: if withouth `{` `}`

BenWhitehead:
Mesos already provides a primitive for killing tasks.

From the scheduler you can call `driver.killTask(taskID)` the executor will then receive the `taskId` via `io.mesosphere.mesos.frameworks.cassandra.CassandraExecutor#killTask` where we can then kill the server.

I think the improved state tracking of the server is good, but we won't need the custom task type to shutdown the server. I think you can leverage this by moving the `case CASSANDRA_SERVER_SHUTDOWN` branch in the executor from `launchTask` to `killTask` and everything should be the same.

You will probably have to add a method to `CassandraCluster` to the effect of `getTaskIDsToKill(offer)` and then call that before calling `getTasksForOffer()` from `io.mesosphere.mesos.frameworks.cassandra.CassandraScheduler#evaluateOffer`

BenWhitehead:
several if without `{` `}` please add braces to all control structures.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: Damn - have read the JDK source wrong (thought it already does a `kill -9`). Java8 has a new `Process.destroyForcibly()` method for that. But we should try to do a `destroy()` before `kill -9` if JMX doesn't work for us.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
snazy: Generally yes - but that way the scheduler immediately knows that the server is dead.
But you're right - I forgot some code to actually update the current health check when it receives that the task is no longer running.

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
BenWhitehead: Sure that sounds good to try destroy before kill -9.  I think they're just stages of trying to make sure it's shutdown.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: Noted a TODO for later.

snazy:
Okay - will exchange `CASSANDRA_SERVER_SHUTDOWN` with `killTask` (didn't knew that) - feels better. Will to the outstanding review changes tomorrow.

snazy:
Pushed rebased branch with review issues fixed.
Also changed how the C* daemon is started:
* removed `-p cassandra.pid` because it's not written when `-f` is given
* takes PID from `java.lang.UNIXProcess`
* also passes PID to scheduler
* honors `LaunchNodeException` and returns `TASK_ERROR`
* checks if process stays alive for at least 500ms and returns `TASK_ERROR` if not (might occur if some settings prevent C* to start)

cassandra-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraExecutor.java
```diff
@@ -223,6 +238,49 @@ public void killTask(final ExecutorDriver driver, final TaskID taskId) {
         if (LOGGER.isDebugEnabled()) {
             LOGGER.debug("killTask(driver : {}, taskId : {})", driver, protoToString(taskId));
         }
+
+        if (serverTask != null && serverTask.getTaskId().equals(taskId)) {
+            killCassandraDaemon(driver);
+        }
+    }
+
+    private void killCassandraDaemon(ExecutorDriver driver) {
```
BenWhitehead: Please move this method down with the other private methods.

BenWhitehead:
I'm not crazy about the UNIXProcess reflection stuff, but it's probably okay until we update to Java 1.8.

BenWhitehead:
One nit about code structure.  Then should be ready to merge after a rebase.

snazy:
Rebased

PR #47, opened 2015-03-11
GossipingPropertyFile snitch

snazy:
Re-add property file snitch (code from my old "003-foo" branches)

Can be merged before PR #44 or #45 (or even after these)

The actual snitch can be configured with code provided by PR #44

### DIFF
--- a/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
+++ b/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
@@ -18,12 +18,13 @@ option java_outer_classname = "CassandraFrameworkProtos";
 message CassandraFrameworkConfiguration {
     optional string frameworkId = 1;
     optional string frameworkName = 2;
-    optional int64 healthCheckIntervalSeconds = 8;
-    optional int64 bootstrapGraceTimeSeconds = 10;
-    repeated PortMapping portMapping = 11;
+    optional int64 healthCheckIntervalSeconds = 3;
+    optional int64 bootstrapGraceTimeSeconds = 4;
+    repeated PortMapping portMapping = 5;
+    optional string snitch = 6;
 
     // default configuration
-    required CassandraConfigRole defaultConfigRole = 6;
+    required CassandraConfigRole defaultConfigRole = 7;
 }
 
 // CassandraConfigRole is currently only used once. But different nodes may require different sizings -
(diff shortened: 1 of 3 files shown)


cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -114,6 +114,12 @@ message CassandraNode {
     required bool seed = 9;
     optional int64 lastRepairTimestamp = 10;
     optional int64 lastCleanupTimestamp = 11;
+
+    optional NodeLocation location = 12;
+}
+message NodeLocation {
+    required string datacenter = 1;
+    required string rack = 2;
```
BenWhitehead: What's the plan on how to populate this info?

snazy: Basically in several ways:
* when you initially setup a cluster pass in the DC name (we don't have racks yet with Mesos)
* when you add new nodes also pass in the DC name (to setup a new DC or to scale-out a DC)

It's also necessary to pass that information, if you want to manage each DC as one Cassandra-framework and connect those DCs using some additional (not fully thought yet).

But as I'm writing these lines, I see your point. Damn. Please hold back this PR for now. Need to think a bit more on this one.

snazy: IMO it makes sense to merge this PR after #44 (due to the config-roles) after some changes - i.e. add the `NodeLocation` to `CassandraConfigRole`.

Meaning of this PR is just to use the `GossipingPropertyFileSnitch` instead of `SimpleSnitch` for on-prem installations. Although that does not help with EC2 + GCE since these snitches take the zone + region names for DC/rack. This gets a bit more complicated if DSE comes into play - but that's a different story.

snazy: Hm - for on-prem installs DC + rack should be injected as Mesos node properties.
And maybe it's not a bad idea to use one framework instance per DC - otherwise one single scheduler would be responsible for all nodes all over the world, which is probably not a good idea. We should then add some inter-scheduler mechanism to exchange seed nodes. But since it's not something that changes often, it's might be fine to do that with some tool called "human" once.

BenWhitehead:
#44 is merged, please rebase and add a config property so that a snitch can be provided at bootstrap time (this will allow us to set EC2 or GCE snitches).

snazy:
Rebased

BenWhitehead:
needs rebase after merging #45

snazy:
Rebased

PR #48, opened 2015-03-11
bounded health check history

snazy:
Limit health check history to N _different_ health check results.
That is - only changes of the `HealthCheckDetails` and `NodeInfo` will be recorded (plus _first_ and _last_ timestamp of occurence).
The `uptimeMillis` field is ignored in change-detection.

TBH I think we should use separate objects in state for each node/executor. But we should do that in a separate PR. We can also merge HC history into `CassandraNode` since `CassandraNode` is also updated on every HC.

This PR is independent from other PRs.

### DIFF
--- a/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
+++ b/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
@@ -47,8 +47,8 @@ message CassandraClusterHealthCheckHistory {
 
 message HealthCheckHistoryEntry {
     required string executorId = 1;
-    required int64 timestampFirst = 2;
-    required int64 timestampLast = 3;
+    required int64 timestampStart = 2;
+    required int64 timestampEnd = 3;
     required HealthCheckDetails details = 4;
 }
 
--- a/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
+++ b/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
@@ -111,11 +111,9 @@ message CassandraNode {
     optional CassandraNodeTask metadataTask = 6;
     optional CassandraNodeTask serverTask = 7;
 
-    repeated HealthCheckDetails healthCheckHistory = 8;
-
-    required bool seed = 9;
-    optional int64 lastRepairTimestamp = 10;
-    optional int64 lastCleanupTimestamp = 11;
+    required bool seed = 8;
+    optional int64 lastRepairTimestamp = 9;
+    optional int64 lastCleanupTimestamp = 10;
 }
 message DataVolume {
     required string path = 1;
--- a/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
+++ b/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
@@ -41,13 +41,15 @@ message CassandraClusterState {
 }
 
 message CassandraClusterHealthCheckHistory {
-    repeated HealthCheckHistoryEntry entries = 1;
+    required int32 maxEntriesPerNode = 1;
+    repeated HealthCheckHistoryEntry entries = 2;
 }
 
 message HealthCheckHistoryEntry {
     required string executorId = 1;
-    required int64 timestamp = 2;
-    required HealthCheckDetails details = 3;
+    required int64 timestampFirst = 2;
+    required int64 timestampLast = 3;
+    required HealthCheckDetails details = 4;
 }
 
 enum ClusterJobType {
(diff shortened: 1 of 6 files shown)


cassandra-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/PersistedCassandraClusterHealthCheckHistoryTest.java
```diff
@@ -0,0 +1,412 @@
...
+    @Test
+    public void testObjEquals() throws Exception {
+        assertTrue(PersistedCassandraClusterHealthCheckHistory.objEquals(null, null));
+        assertFalse(PersistedCassandraClusterHealthCheckHistory.objEquals(null, ""));
+        assertFalse(PersistedCassandraClusterHealthCheckHistory.objEquals("", null));
+        assertTrue(PersistedCassandraClusterHealthCheckHistory.objEquals("1", "1"));
+        assertTrue(PersistedCassandraClusterHealthCheckHistory.objEquals(1, 1));
+        assertFalse(PersistedCassandraClusterHealthCheckHistory.objEquals("1", "2"));
+        assertFalse(PersistedCassandraClusterHealthCheckHistory.objEquals(1, 2));
+    }
```
BenWhitehead: I have another test (below) to add that fails.

Mesos framework messages aren't guaranteed to be delivered in order, so the logic in `record` will need to be updated to account for this fact.

```java
    @Test
    public void testAllOscillatingOutOfOrder() throws Exception {
        State state = new InMemoryState();
        PersistedCassandraClusterHealthCheckHistory hcHistory = new PersistedCassandraClusterHealthCheckHistory(state);

        hcHistory.record("abc", 10, unhealthy());
        hcHistory.record("abc", 11,   healthy());
        hcHistory.record("abc", 13, unhealthy());
        hcHistory.record("abc", 16, unhealthy());
        hcHistory.record("abc", 14,   healthy());
        hcHistory.record("abc", 19, unhealthy());
        hcHistory.record("abc", 15,   healthy());
        hcHistory.record("abc", 12,   healthy());
        hcHistory.record("abc", 25, unhealthy());
        hcHistory.record("abc", 17,   healthy());
        hcHistory.record("abc", 20,   healthy());
        hcHistory.record("abc", 21,   healthy());
        hcHistory.record("abc", 22, unhealthy());
        hcHistory.record("abc", 24,   healthy());
        hcHistory.record("abc", 18,   healthy());
        hcHistory.record("abc", 23,   healthy());

        final CassandraClusterHealthCheckHistory history = hcHistory.get();
        final List<CassandraFrameworkProtos.HealthCheckHistoryEntry> list = history.getEntriesList();

        CassandraFrameworkProtos.HealthCheckHistoryEntry prev = null;
        for (final CassandraFrameworkProtos.HealthCheckHistoryEntry entry : list) {
            if (prev != null) {
                assertThat(entry.getTimestampFirst()).isLessThanOrEqualTo(entry.getTimestampLast());
                assertThat(prev.getTimestampLast()).isLessThanOrEqualTo(entry.getTimestampFirst());
            }
            prev = entry;
        }
    }

    static HealthCheckDetails healthy() {
        return healthCheckDetails(true);
    }
    static HealthCheckDetails unhealthy() {
        return healthCheckDetails(false);
    }

    static HealthCheckDetails healthCheckDetails(final boolean healthy) {
        return HealthCheckDetails.newBuilder()
            .setHealthy(healthy)
            .setInfo(CassandraFrameworkProtos.NodeInfo.newBuilder()
                .setClusterName("cluster")
                .setDataCenter("dc1")
                .setUptimeMillis(1).build()).build();
    }
```

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -41,13 +41,15 @@ message CassandraClusterState {
...
-    repeated HealthCheckHistoryEntry entries = 1;
+    required int32 maxEntriesPerNode = 1;
+    repeated HealthCheckHistoryEntry entries = 2;
 }
 
 message HealthCheckHistoryEntry {
     required string executorId = 1;
-    required int64 timestamp = 2;
-    required HealthCheckDetails details = 3;
+    required int64 timestampFirst = 2;
```
BenWhitehead: Since this object is representing a time range lets use `start` and `end` rather than `first` and `last`.

snazy: done

cassandra-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/PersistedCassandraClusterHealthCheckHistoryTest.java
snazy: Damn out-of-order protocols ;)

In this case - I've added some code to silently drop records that are older that the most-recent one. Since we're not building a descent monitoring tool it should be fine IMO to drop outdated health-check info.

snazy:
Rebased branch against `rewrite`

BenWhitehead:
thanks

PR #49, opened 2015-03-11
NIT: test deps with correct scope

snazy:


BenWhitehead:
Test scope is already set on these at the dependencyManagement level see [here](https://github.com/mesosphere/cassandra-mesos/blob/rewrite/pom.xml#L196)

PR #50, opened 2015-03-11
Add API to inquire live nodes by clients

snazy:
Adds an API endpoint to produce a list of live nodes for specific tools or client libraries.

This is a prerequisite for stress testing. It provides connect information for `nodetool`, `cassandra-stress`, `cqlsh` plus general-purpose JSON and plain text output.

General invocation:
`http://host:port/live-endpoints/json/2`   retrieves two live nodes via json
`http://host:port/live-endpoints/cqlsh`   retrieves one live nodes as a argument (list) for cqlsh

Available 'applications' are `json`, `text`, `cqlsh`, `stress`, `nodetool`

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -233,3 +233,58 @@ JRE_FILE_PATH=${PROJECT_DIR}/target/framework-package/jdk.tar.gz
 CASSANDRA_FILE_PATH=${PROJECT_DIR}/target/framework-package/cassandra.tar.gz
 
 ```
+
+## Using Cassandra tools
+
+You can use standard command line tools delivered with Apache Cassandra against clusters running on
+Apache Mesos.
+
+This framework provides API endpoints for most tools. All you need is `curl` and the hostname/IP of
+the node running the scheduler of this framework.
+
+An example how to invoke `nodetool`:
+
+```
+cassandra/2.1/bin/nodetool `curl -s http://192.168.5.101:18080/live-nodes/nodetool` status
+Datacenter: datacenter1
+=======================
+Status=Up/Down
+|/ State=Normal/Leaving/Joining/Moving
+--  Address    Load       Tokens  Owns (effective)  Host ID                               Rack
+UN  127.0.0.1  41,12 KB   256     100,0%            dc04253f-d878-4d44-acf3-dc014e058b03  rack1
+UN  127.0.0.2  55,51 KB   256     100,0%            685e82e9-fa26-4d12-bee3-13d4f823f5c9  rack1
+```
+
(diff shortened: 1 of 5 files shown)


cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -254,6 +256,94 @@ private static void writeTask(JsonGenerator json, String name, CassandraFramewor
...
 
+    @GET
+    @Path("/live-nodes/{forTool}")
+    public Response liveEndpoints(@PathParam("forTool") String forTool) {
+        return liveEndpoints(forTool, 1);
+    }
+
+    @GET
+    @Path("/live-nodes/{forTool}/{nodeCount}")
+    public Response liveEndpoints(@PathParam("forTool") String forTool, @PathParam("nodeCount") Integer nodeCount) {
```
BenWhitehead: I would prefer each of these scenarios be split out into individual methods.

I think we should switch to using query params for limit rather than path element. All URL methods could specify a `limit` query param `@QueryParam("limit") @DefaultValue("3")` where we can set a default limit if one isn't specified.

In terms of URLs I think we should tweak to be (with target mime type)
* `/live-nodes` -> `application/json`
* `/live-nodes/tool/cqlsh` -> `text/x-apache-cassandra-cqlsh`
* `/live-nodes/tool/stress` -> `text/x-apache-cassandra-stress`
* `/live-nodes/tool/nodetool` -> `text/x-apache-cassandra-nodetool`

I'm having a hard time imagining a case where `ascii` would be useful. If you don't have a use case in mind please remove it.

README.markdown
```diff
@@ -196,3 +196,58 @@ JRE_FILE_PATH=${PROJECT_DIR}/target/framework-package/jdk.tar.gz
...
+curl -s http://192.168.5.101:18080/live-nodes/json/2
+{
+  "nativePort" : 9042,
+  "rpcPort" : 9160,
+  "jmxPort" : 7199,
+  "liveNodes" : [ "127.0.0.2", "127.0.0.1" ]
+}
+```
+
+Example for `ascii` endpoint:
```
BenWhitehead: I'm having a hard time imagining a case where ascii would be useful. If you don't have a use case in mind please remove it.

snazy:
pure text can be useful for simple shell scripts - much easier than parsing JSON in bash for example.

snazy:
Updated to use query-param and other endpoints & rebased

BenWhitehead:
Needs rebase after merging #45

snazy:
OK, rebased

PR #51, opened 2015-03-13
Update to maven-protoc-plugin 

BenWhitehead:
New version includes fix for breaking changes to maven-toolchains-plugin introduced in maven 3.2.5.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,7 +50,7 @@
     </issueManagement>
 
     <prerequisites>
-        <maven>3.0.0</maven>
+        <maven>3.2.0</maven>
     </prerequisites>
 
     <properties>
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,10 @@
         <url>https://github.com/mesosphere/cassandra-mesos/issues</url>
     </issueManagement>
 
+    <prerequisites>
+        <maven>3.0.0</maven>
+    </prerequisites>
+
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <version.java>1.7</version.java>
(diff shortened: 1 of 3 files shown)


BenWhitehead:
@snazy Please review.

pom.xml
```diff
@@ -49,6 +49,10 @@
         <url>https://github.com/mesosphere/cassandra-mesos/issues</url>
     </issueManagement>
 
+    <prerequisites>
+        <maven>3.0.0</maven>
+    </prerequisites>
```
snazy: Overall LGTM - except that I'd like to see Maven 3.2 as the prerequisite. Older versions (3.1.1 + 3.0.5) produce unnecessary warnings. So - if `3.0.0` is changed to `3.2.0` I'm +1 on this PR.

```
[WARNING] 'dependencyManagement.dependencies.dependency.exclusions.exclusion.groupId' for org.apache.cassandra:cassandra-all:jar with value '*' does not match a valid id pattern. @ io.mesosphere.mesos.frameworks:cassandra:0.1.0-SNAPSHOT, /Users/snazy/devel/mesos/cassandra-mesos/pom.xml, line 144, column 34
[WARNING] 'dependencyManagement.dependencies.dependency.exclusions.exclusion.artifactId' for org.apache.cassandra:cassandra-all:jar with value '*' does not match a valid id pattern. @ io.mesosphere.mesos.frameworks:cassandra:0.1.0-SNAPSHOT, /Users/snazy/devel/mesos/cassandra-mesos/pom.xml, line 145, column 37
```

snazy:
Just a (minor) prerequisite thing - otherwise LGTM

PR #52, opened 2015-03-13
Replace node

snazy:
Now that's #45 is merged, here's the follow-up for "replace node".

To replace a node, it must first be terminated (i.e. executor shutdown).

### DIFF
--- a/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
+++ b/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
@@ -97,6 +97,8 @@ message CassandraClusterState {
     repeated CassandraNode nodes = 1;
     repeated ExecutorMetadata executorMetadata = 2;
     optional int64 lastServerLaunchTimestamp = 3;
+    repeated string replaceNodeIps = 4;
+    required int32 nodesToAcquire = 5;
 }
 
 message CassandraClusterHealthCheckHistory {
@@ -178,6 +180,8 @@ message CassandraNode {
 
     optional int32 cassandraDaemonPid = 12;
 
+    optional string replacementForIp = 13;
+
     enum TargetRunState {
         // If the server task is not active, it is started.
         // Note that general bootstrap-grace-time settings and other checks (e.g. currently other nodes joining) apply.
@@ -188,9 +192,11 @@ message CassandraNode {
         // If the server-task is active, a server-task shutdown is initiated.
         // Note that general bootstrap-grace-time settings and other checks (e.g. currently other nodes joining) apply.
         RESTART = 2;
+        // Terminate all tasks including the executor itself. There's no way back for a terminated node.
+        TERMINATE = 3;
     }
 
-    optional NodeLocation location = 13;
+    optional NodeLocation location = 14;
 }
 message NodeLocation {
     required string datacenter = 1;
--- a/cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
+++ b/cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
@@ -695,7 +695,7 @@ public final class ApiController {
         return nodeStatusUpdate(cassandraNode);
     }
 
-    @GET
+    @POST
     @Path("/node/terminate/{node}")
     public Response nodeTerminate(@PathParam("node") String node) {
         CassandraFrameworkProtos.CassandraNode cassandraNode = cluster.nodeTerminate(node);
@@ -735,7 +735,7 @@ public final class ApiController {
         return Response.status(status).entity(sw.toString()).type("application/json").build();
     }
 
-    @GET
+    @POST
     @Path("/node/replace/{node}")
     public Response nodeReplace(@PathParam("node") String node) {
         int status = 200;
--- a/cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
+++ b/cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
@@ -179,9 +179,22 @@ public final class ApiController {
 
             json.writeStartObject();
             CassandraFrameworkProtos.CassandraClusterState clusterState = cluster.getClusterState().get();
+
+            json.writeArrayFieldStart("replaceNodes");
+            for (String ip : clusterState.getReplaceNodeIpsList()) {
+                json.writeString(ip);
+            }
+            json.writeEndArray();
+
+            json.writeNumberField("nodesToAcquire", clusterState.getNodesToAcquire());
+
             json.writeArrayFieldStart("nodes");
             for (CassandraFrameworkProtos.CassandraNode cassandraNode : clusterState.getNodesList()) {
                 json.writeStartObject();
+
+                if (cassandraNode.hasReplacementForIp()) {
+                    json.writeStringField("replacementForIp", cassandraNode.getReplacementForIp());
+                }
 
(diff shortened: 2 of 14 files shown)


cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -178,6 +180,8 @@ message CassandraNode {
 
     optional int32 cassandraDaemonPid = 12;
 
+    optional string replacementForIp = 13;
```
BenWhitehead: Is this something that will be tracked long term? Do we need to know about the replaced ip after the replacement has taken place?

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -682,6 +695,14 @@ public Response nodeRestart(@PathParam("node") String node) {
         return nodeStatusUpdate(cassandraNode);
     }
 
+    @GET
```
BenWhitehead: This should be a `POST`

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -713,4 +734,44 @@ private static Response nodeStatusUpdate(CassandraFrameworkProtos.CassandraNode
         }
         return Response.status(status).entity(sw.toString()).type("application/json").build();
     }
+
+    @GET
```
BenWhitehead: This should be a `POST`

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -713,4 +734,44 @@ private static Response nodeStatusUpdate(CassandraFrameworkProtos.CassandraNode
...
+    @Path("/node/replace/{node}")
+    public Response nodeReplace(@PathParam("node") String node) {
```
BenWhitehead: What is the use case you are imagining this endpoint satisfies? I thought the scheduler was responsible for making decisions about when a node should be replaced, this seems like we're making it the operators responsibility.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -306,23 +318,31 @@ public TasksForOffer getTasksForOffer(@NotNull final Protos.Offer offer) {
...
-                        newNode
-                ));
+                CassandraNode newNode = buildCassandraNode(offer, buildSeedNode, replacementForIp);
+                clusterState.nodeAcquired(newNode);
                 node = CassandraNode.newBuilder(newNode);
             } else
                 node = CassandraNode.newBuilder(nodeOption.get());
 
             if (!node.hasCassandraNodeExecutor()) {
+                if (node.getTargetRunState() == CassandraNode.TargetRunState.TERMINATE) {
```
BenWhitehead: Do terminated nodes stick around in the state info after we've verified that the node has terminated?  It seems like we're checking the terminated state a lot.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -977,4 +1039,30 @@ private boolean isLiveNode(CassandraNode node) {
             return false;
         return info.getNativeTransportRunning() && info.getRpcServerRunning();
     }
+
+    public CassandraNode replaceNode(String node) throws ReplaceNodePreconditionFailed {
```
BenWhitehead: Please add tests that check both the positive & negative cases of each of these branches. There is a lot of logic here that could go wrong if we don't account for something correctly.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -97,6 +97,8 @@ message CassandraClusterState {
     repeated CassandraNode nodes = 1;
     repeated ExecutorMetadata executorMetadata = 2;
     optional int64 lastServerLaunchTimestamp = 3;
+    repeated string replaceNodeIps = 4;
+    required int32 nodesToAcquire = 5;
```
BenWhitehead: How does this differ from `numberOfNodes` in `CassandraConfigRole`? It seems like the same thing to me.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
BenWhitehead: Code coverage info can be seen here: https://teamcity.mesosphere.io/viewLog.html?buildId=12455&tab=coverage_idea&buildTypeId=Oss_Mesos_Cassandra_CassandraFramework

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: It's not tracked long-term. Once the node has been replaced, the field is cleared.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
snazy: fixed

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
snazy: fixed

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
snazy: Nope - the operator's responsible to replace a node. It's nothing that can or should be done automatically.

Of course, there are multiple scenarios:
* manual, planned node replacement: for example when you scale up one node after another (replacing with "better" hardware)
* determining that a node is no longer available (specific to Cassandra-on-Mesos): But when exactly is a node "no longer available"?

We should be very careful about doing such things automatically.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: Currently, yes. We should "lock out" the terminated and replaced node for some days until the last node in the cluster has forgotten the replaced node. (It may take some time until the last node recognizes that - such a "last" node could be one that has been dead for some time, too.)

Sure, we re-check such stuff as soon as we have permanently allocated resources from Mesos (get the same data volumes and ports from Mesos after a restart of the executor)

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: That's the "current" number of new nodes (executors) to acquire. It's decremented once an executor is acquired.
Initially it's equal to `numberOfNodes`. But once you call `replaceNode` is incremented.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: k, added more tests

PR #53, opened 2015-03-13
Tools using live endpoint API

snazy:
JavaDriver extension using Cassandra-on-Mesos framework API endpoint to configure native protocol port and initial contact points.

It's implemented as a project that's independent from cassandra-on-mesos-framework. At the moment it's basically just a demo how it could work. For a more complete set we could add code for more drivers (e.g. python) and more convenient shell scripts instead of `nodetool curl`

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -325,3 +325,30 @@ curl -s http://192.168.5.101:18080/live-nodes/ascii/2
 ```
 
 Note that the implementation does a best-effort approach to return random nodes.
+
+## Using Cassandra-on-Mesos with DataStax Java Driver
+
+The artifact `java-driver-extension` provides the `CassandraOnMesos` tool class that 
+configures the DataStax Java Driver with native protocol port and contact points from the information
+available from the Cassandra-on-Mesos framework.
+
+You need to have the base-API-URI provided the framework.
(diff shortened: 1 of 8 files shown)


BenWhitehead:
Good idea to improve tooling support. I'm going to hold off reviewing in detail for now since testing (and changes related to it) is higher priority.

snazy:
Include this PR for 0.1?
People could use it to start experimenting with it.

BenWhitehead:
Not for the release we're trying to get out for monday. 0.1 isn't going to be released by monday, just an Early Access build

snazy:
BTW: Don't you have a public holiday on Monday (Easter) ?

Here we have a public holiday today and on Monday - but I don't care about that.


> Am 03.04.2015 um 20:37 schrieb BenWhitehead <notifications@github.com>:
> 
> Not for the release we're trying to get out for monday. 0.1 isn't going to be released by monday, just an Early Access build
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/mesosphere/cassandra-mesos/pull/53#issuecomment-89385232>.
> 

—
Robert Stupp
@snazy

BenWhitehead:
nope no public holiday on monday for us.

PR #54, opened 2015-03-16
remove superfluous src

snazy:


BenWhitehead:
This package-info.java is necessary for the [idea-maven-plugin](https://github.com/mesosphere/cassandra-mesos/blob/rewrite/pom.xml#L285) that adds assertions into the code for `@NotNull`.  I didn't notice that it had been disabled, we should re-enable it and this file will then be necessary.

snazy:
Hm - that's ideauidesigner-maven-plugin (_plugin provides the capability to convert *.form files into Java code_).

BenWhitehead:
You are correct that it supports compiling *.form files, but the same plugin runs the javac2 compiler which adds the Assertions for `@NotNull`.

snazy:
Heh - can't believe that - but you're right. Crazy that a UI-plugin does such things...
The size of the class files differs with and without that plugin.
Will revert the meaning of this PR. Give me a few minutes.

PR #55, opened 2015-03-16
Bind API http server to any inet address

snazy:


BenWhitehead:
What's the advantage of doing this?  We still have to resolve an IP so that it can be sent to the slave to pull down the resources.

snazy:
It's for other tools (possibly on an admin's PC) and multi-homed systems. But we can make it configurable, if you want. E.g. some want it on `0.0.0.0` others on `1.2.3.4` and so on.

PR #62, opened 2015-03-17
Ability to change seed status of nodes + push new config to existing nodes

snazy:
Without being able to change the `seed` flag of a node, it's not possible to replace a seed node.

This PR adds the ability to change the seed node status.

It also adds the ability to push new configuration files without restarting the Cassandra process.

Note: it's not permitted to make the last live seed node a non-seed node.


snazy:
Rebased branch

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -271,11 +279,19 @@ message CassandraServerRunTask {
...
-    required JmxConnect jmx = 6;
+    required CassandraServerConfig cassandraServerConfig = 4;
+    required JmxConnect jmx = 5;
+}
+
+message UpdateConfigTask {
+    required CassandraServerConfig cassandraServerConfig = 1;
+}
+
+message CassandraServerConfig {
```
BenWhitehead: Should the `TaskEnv` object also be moved into here? It contains config changes for cassandra-env.sh.

snazy: Hm - you're right. Moved it.

PR #57, opened 2015-03-16
Qa report bundle

snazy:
This PR implements a new API endpoint (`qaReportResources`) that can be used to get and pull all necessary information for a complete QA report.

There are two variants of the endpoint: one returning JSON and one returning plain text. The second one is to be used by cluster-loadtest to also include the C* node logfiles.

Will provide some tooling on that endpoint in a future PR - it's basically "raw" for cluster-loadtest.

PR #63, opened 2015-03-19
Cluster loadtest

snazy:
Load testing against Cassandra-on-Mesos.
Depends on #57 

Only commit specific to this PR is the one called "load testing"


BenWhitehead:
Can you add a readme into the cluster-loadtest directory with a quick explanation about what the test runs and what to look for in the output?

snazy:
Added a short readme.

BenWhitehead:
Thanks

PR #64, opened 2015-03-19
protobuf model and code cleanup

snazy:
Add documentation to model.proto,
    add clusterName for framework config,
    extract resources (cpu, mem, disk, ports) to separate message,
    remove unused CleanupScheduledJob + RepairScheduledJob,
    rename URI to FileDownload,
    clarified different TaskType enums used nearby each other

This PR depends on #62 - only commit 5937cb7 is specific to this PR.

I'd really like to remove the datacenter, rack and NodeLocation stuff for now until we have a properly designed solution. WDYT?

### DIFF
--- a/cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
+++ b/cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
@@ -96,11 +96,11 @@ public final class Main {
         final double    resourceCpuCores        = Double.parseDouble(   Env.option("CASSANDRA_RESOURCE_CPU_CORES").or("2.0"));
         final long      resourceMemoryMegabytes = Long.parseLong(       Env.option("CASSANDRA_RESOURCE_MEM_MB").or("2048"));
         final long      resourceDiskMegabytes   = Long.parseLong(       Env.option("CASSANDRA_RESOURCE_DISK_MB").or("2048"));
+        final long      javaHeapMb              = Long.parseLong(       Env.option("CASSANDRA_RESOURCE_HEAP_MB").or("0"));
         final long      healthCheckIntervalSec  = Long.parseLong(       Env.option("CASSANDRA_HEALTH_CHECK_INTERVAL_SECONDS").or("60"));
         final long      bootstrapGraceTimeSec   = Long.parseLong(       Env.option("CASSANDRA_BOOTSTRAP_GRACE_TIME_SECONDS").or("120"));
         final String    cassandraVersion        =                       "2.1.2"; // TODO Env.option("CASSANDRA_VERSION").or("2.1.2");
-        final String    clusterName             = frameworkName(        Env.option("CASSANDRA_CLUSTER_NAME"));
-        final String    frameworkName           = frameworkName(        Env.option("FRAMEWORK_NAME").or(Env.option("CASSANDRA_CLUSTER_NAME")));
+        final String    frameworkName           = frameworkName(        Env.option("FRAMEWORK_NAME"));
         final String    zkUrl                   =                       Env.option("CASSANDRA_ZK").or("zk://localhost:2181/cassandra-mesos");
         final long      zkTimeoutMs             = Long.parseLong(       Env.option("CASSANDRA_ZK_TIMEOUT_MS").or("10000"));
         final String    mesosMasterZkUrl        =                       Env.option("MESOS_ZK").or("zk://localhost:2181/mesos");
@@ -120,22 +120,19 @@ public final class Main {
             throw new IllegalArgumentException("number of nodes (" + executorCount + ") and/or number of seeds (" + seedCount + ") invalid");
         }
 
-        CassandraFrameworkProtos.CassandraConfigRole.Builder defaultConfigRole = CassandraFrameworkProtos.CassandraConfigRole.newBuilder()
-            .setCassandraVersion(cassandraVersion)
-            .setResources(CassandraFrameworkProtos.TaskResources.newBuilder()
-                .setCpuCores(resourceCpuCores)
-                .setDiskMb(resourceDiskMegabytes)
-                .setMemMb(resourceMemoryMegabytes))
-            .setNumberOfNodes(executorCount)
-            .setNumberOfSeeds(seedCount)
-            .setMesosRole(mesosRole);
         final PersistedCassandraFrameworkConfiguration configuration = new PersistedCassandraFrameworkConfiguration(
             state,
             frameworkName,
-            clusterName,
-            defaultConfigRole.build(),
             healthCheckIntervalSec,
-            bootstrapGraceTimeSec
+            bootstrapGraceTimeSec,
+            cassandraVersion,
+            resourceCpuCores,
+            resourceDiskMegabytes,
+            resourceMemoryMegabytes,
+            javaHeapMb,
+            executorCount,
+            seedCount,
+            mesosRole
         );
 
 
@@ -159,7 +156,7 @@ public final class Main {
             clock,
             httpServerBaseUri.toString(),
             new ExecutorCounter(state, 0L),
-            new PersistedCassandraClusterState(state, defaultConfigRole.build()),
+            new PersistedCassandraClusterState(state, executorCount, seedCount),
             new PersistedCassandraClusterHealthCheckHistory(state),
             new PersistedCassandraClusterJobs(state),
             configuration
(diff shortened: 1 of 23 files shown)


cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
```diff
@@ -99,7 +99,8 @@ private static int _main() throws UnknownHostException {
         final long      healthCheckIntervalSec  = Long.parseLong(       Env.option("CASSANDRA_HEALTH_CHECK_INTERVAL_SECONDS").or("60"));
         final long      bootstrapGraceTimeSec   = Long.parseLong(       Env.option("CASSANDRA_BOOTSTRAP_GRACE_TIME_SECONDS").or("120"));
         final String    cassandraVersion        =                       "2.1.2"; // TODO Env.option("CASSANDRA_VERSION").or("2.1.2");
-        final String    frameworkName           = frameworkName(        Env.option("CASSANDRA_CLUSTER_NAME"));
+        final String    clusterName             = frameworkName(        Env.option("CASSANDRA_CLUSTER_NAME"));
+        final String    frameworkName           = frameworkName(        Env.option("FRAMEWORK_NAME").or(Env.option("CASSANDRA_CLUSTER_NAME")));
```
BenWhitehead: We don't need the framework name and the zk url manipulation. The zk URL that is provided allows for nesting of nodes if the user wants to do that. See #68 where I added several tests to illustrate the supported urls and formats.

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
```diff
@@ -161,7 +163,7 @@ private static int _main() throws UnknownHostException {
             clock,
             httpServerBaseUri.toString(),
             new ExecutorCounter(state, 0L),
-            new PersistedCassandraClusterState(state, defaultConfigRole),
+            new PersistedCassandraClusterState(state, defaultConfigRole.build()),
```
BenWhitehead: Here we should pass the seed count and node count rather than passing a big object that has the values pulled out of it.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/ApiController.java
```diff
@@ -52,6 +54,7 @@ public Response indexPage(@Context UriInfo uriInfo) {
             String baseUrl = uriInfo.getBaseUri().toString();
 
             json.writeStringField("configuration", baseUrl + "config");
+            json.writeStringField("seedNodes", baseUrl + "seed-nodes");
             json.writeStringField("allNodes", baseUrl + "nodes");
```
BenWhitehead: It would be good to add the new URLs for tools and things that have been added.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -15,44 +15,95 @@ package io.mesosphere.mesos.frameworks.cassandra;
 
 option java_outer_classname = "CassandraFrameworkProtos";
 
+/**
+ * Base configuration object.
+ * This is a top-level object in the state hierarchy.
+ */
 message CassandraFrameworkConfiguration {
+    /**
+     * ID of the framework. Assigned by Mesos when the scheduler is registered.
```
BenWhitehead: More accurate comment: 
>  ID of the framework. 
>  
>  Upon initial framework registration Mesos will generate the framework id and return it as part of `Scheduler.registered`
>  The framework will use this value (if set) when it restarts and re-registers with Mesos

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -15,44 +15,95 @@ package io.mesosphere.mesos.frameworks.cassandra;
...
+     * still have the same Cassandra cluster name.
+     */
+    required string frameworkName = 2;
+    /**
+     * Name of the Cassandra cluster.
+     * This is different from frameworkName since if you have setup multiple Cassandra-on-Mesos
+     * frameworks for each logical data center (with different frameworkName and frameworkId) you
+     * still have the same Cassandra cluster name.
+     */
+    required string clusterName = 3;
```
BenWhitehead: Part of multi-dc that we're deferring until later. Please remove.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -83,36 +134,89 @@ message CassandraConfigRole {
...
+ * Current status of the cluster.
+ * This is a top-level object in the state hierarchy.
+ */
 message CassandraClusterState {
+    /**
+     * List of all known nodes.
+     */
     repeated CassandraNode nodes = 1;
+    /**
+     * List of all known executors.
```
BenWhitehead: > List of Metadata collected during executor startup

Actual executor tracking is done at the node level.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -83,36 +134,89 @@ message CassandraConfigRole {
...
+     * option passed to the Cassandra process.
+     */
     required int32 nodesToAcquire = 5;
     required int32 seedsToAcquire = 6;
 }
 
+/**
+ * Contains a history of the last health-checks that were received from all nodes.
+ * The implementation should not add a new entry when it is similar to the previous one.
+ * Instead it should update the timespan in the previous one.
```
BenWhitehead: We should move:
> The implementation should not add a new entry when it is similar to the previous one. Instead it should update the timespan in the previous one.

To the "Persistent" class that implements this logic. The model object doesn't care.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -83,36 +134,89 @@ message CassandraConfigRole {
...
     required int32 maxEntriesPerNode = 1;
+    /**
+     * Health check history entries.
+     */
     repeated HealthCheckHistoryEntry entries = 2;
 }
 
+/**
+ * Health check history entry as part of CassandraClusterHealthCheckHistory message.
+ */
```
BenWhitehead: Potential new phrasing:

> Represents and time span where the specified executor was in a particular state.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -286,19 +634,44 @@ message UpdateConfigTask {
...
 
+/**
+ * Information how to connect to Cassandra's JMX server.
+ */
 message JmxConnect {
+    /**
+     * The TCP port the JMX server uses. Usually this port equals to the default port in
+     * CassandraFrameworkConfiguration.portMapping. In a development environment where all nodes are running on
+     * the same machine, the JMX port differs per node (thanks to Sun/Oracle, who didn't made it possible to
+     * bind JMX server to an IP address).
```
BenWhitehead: Simplify to:
> The TCP port the JMX server uses. Usually this port equals to the default port in CassandraFrameworkConfiguration.portMapping.

The other details have no bearing on the field itself.

snazy:
updated according to your comments

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
```diff
@@ -112,28 +113,26 @@ private static int _main() throws UnknownHostException {
             matcher.group(1),
             zkTimeoutMs,
             TimeUnit.MILLISECONDS,
-            matcher.group(2)
+            matcher.group(2) + '/' + frameworkName
```
BenWhitehead: Please remove this added sub-directory it's not needed give the user can specify what ever zk node they want as the base.

BenWhitehead:
One small point then ready to merge.

snazy:
removed that (superfluous) thing

PR #65, opened 2015-03-19
Update executor memory and disk sizing

BenWhitehead:
We're setting `-Xms256m -Xmx256m` but weren't claiming the resources so when cgroup isolation is enabled the task is killed with OOM.


BenWhitehead:
@snazy please review.

snazy:
We should increase provisioned memory to a bit more than just the heap size. I think 300 should work (including all JVM overhead).

PR #66, opened 2015-03-20
generate THIRD-PARTY.txt + download licenses

snazy:
files are generated on build of parent pom

If dependencies are changes these files need to be committed, too


BenWhitehead:
What's the motivation for checking the licenses into the repo when they're re-fetched and aggregated  each build?

pom.xml
```diff
@@ -344,6 +344,48 @@
...
+                    <excludedScopes>test</excludedScopes>
+                    <excludedGroups>${project.groupId}</excludedGroups>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>download-licenses</id>
+                        <goals>
+                            <goal>download-licenses</goal>
+                        </goals>
+                        <phase>generate-resources</phase>
```
BenWhitehead: This needs to bind to a phase later than package otherwise the goal will use the "installed" package which usually isn't there. `verify` is probably our best bet since it's post package but pre-install (since there isn't `post-package`)

pom.xml
```diff
@@ -344,6 +344,48 @@
...
+                            <licensesOutputDirectory>licenses</licensesOutputDirectory>
+                        </configuration>
+                    </execution>
+
+                    <execution>
+                        <id>default-cli</id>
+                        <goals>
+                            <goal>aggregate-add-third-party</goal>
+                        </goals>
+                        <configuration>
```
BenWhitehead: Bind to same phase as `download-licenses`

BenWhitehead:
See https://teamcity.mesosphere.io/viewLog.html?buildId=12839&buildTypeId=Oss_Mesos_Cassandra_CassandraFramework&tab=buildLog#_focus=188 for build error related to maven phase.

snazy:
Ah - the build failure happens when there's nothing in the local M2 repo.

Unfortunately Maven always builds the parent project first and then all child modules - even the 'verify' phase happens before the child modules are built.

I worked around that by moving the license-stuff into a separate profile - we usually don't need it to execute during every build. All we need is that THIRD-PARTY.txt file, which should be included in every distribution.

BenWhitehead:
I still ran into issues being able to cleanly build without an install.

The only way I managed to get it to run cleanly every time profile or not was to create another module and have that module depend on all other modules.

Can you please update this change set to add a new module `cassandra-package` (later `cassandra-mesos-package`) that contains the licensing maven plugin.  I really don't want to have to run an install every time just to be able to generate the report.

snazy:
Sure - that's the better option. We could also get rid of `package.bash` with that new module. WDYT?

snazy:
OK, I've updated the PR to:
* added new `cassandra-dist` module that replaces the `package.bash` script and also collects aggregates license stuff
* generate `THIRD-PARTY.txt` + `licenses` directories for executor, framework and dist modules

`cassandra-dist` can also deploy to S3 when invoked with profile `deploy` - though, I could not test it.

BenWhitehead:
I think the idea of packaging in dist is good, but I think we'll need to keep the packaging script since it has logic built into it to allow for snapshots from branches to coexist without clobbering each other. It also has logic built into it to generate a `marathon.json` file which is still needed.

Please remove the downloaded licenses and add the following snippet to `.gitignore` so that they aren't checked into the repo and causing revision churn.

```diff
diff --git a/.gitignore b/.gitignore
index f1f0b62..6443b97 100644
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,8 @@ target/
 
 driver-extensions/cluster-loadtest/cassandra.logdir_IS_UNDEFINED/**
 driver-extensions/cluster-loadtest/loadtest-*/**
+
+# Compiled Licenses
+**/licenses/
+**/THIRD-PARTY.txt
+
```

snazy:
Re-added package.bash, but changed it to use the target dir of the cassandra-dist module.
(Rebased + force-pushed the branch).
Also doesn't attach the big generated .tar.gz file to installable artifacts (it's around 70MB large).

BenWhitehead:
Please remove `cassandra-executor/src/main/licenses` once the third party licenses are removed this is ready to merge.

snazy:
Removed + rebased

BenWhitehead:
Thanks

PR #69, opened 2015-03-22
add missing state change code, add more unit test code

snazy:
* split utests in CassandraSchedulerTests into separate source files
* fix mesos-resource-role matchin (match everything on `*`)
* fix buggy cluster job state handling when executor is lost or task is lost
* add more utests + enhance existing utests (should increase code coverage)
* fix buggy change-seed code

### DIFF
--- a/cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
+++ b/cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
@@ -120,7 +120,7 @@ public final class CassandraCluster {
         }
         Long port = defaultPortMappings.get(name);
         if (port == null) {
-            throw new IllegalStateException("no port mapping for " + name);
+            throw new IllegalArgumentException("no port mapping for " + name);
         }
         return port.intValue();
     }
@@ -190,6 +190,9 @@ public final class CassandraCluster {
                 case SERVER:
                     builder.clearCassandraDaemonPid();
                     break;
+                case CLUSTER_JOB:
+                    jobsState.clearClusterJobCurrentNode(status.getExecutorId().getValue());
+                    break;
             }
 
             newNodes.add(builder.build());
@@ -220,6 +223,7 @@ public final class CassandraCluster {
                 }
             })
             .transform(cassandraNodeBuilderToCassandraNode());
+        jobsState.clearClusterJobCurrentNode(executorId);
         clusterState.nodes(newArrayList(update));
         removeExecutorMetadata(executorId);
     }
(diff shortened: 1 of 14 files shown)


cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -823,13 +827,13 @@ private static CassandraNodeTask getMetadataTask(@NotNull final String executorI
         final Long availableMem = resourceValueLong(headOption(index.get("mem"))).or(0L);
         final Long availableDisk = resourceValueLong(headOption(index.get("disk"))).or(0L);
 
-        if (availableCpus <= cpu) {
-            errors.add(String.format("Not enough cpu resources for role %s. Required %f only %f available.", mesosRole, cpu, availableCpus));
+        if (availableCpus < cpu) {
```
BenWhitehead: Why the change to strictly less than? Won't this now fail if the task needs `2` and the offer has exactly `2`?

cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/CassandraFrameworkProtosUtils.java
```diff
@@ -339,7 +339,7 @@ public ResourceHasExpectedRole(@NotNull final String role) {
 
         @Override
         public boolean apply(final Resource item) {
-            return item.getRole().equals(role);
+            return "*".equals(role) || "*".equals(item.getRole()) || item.getRole().equals(role);
```
BenWhitehead: I don't think we want to absorb the extra logic necessary to be able to combine offers right now to the point that we could use resources across multiple roles.  Please remove this for now, we can decide if we want to do something around multi-role support later on.

BenWhitehead: For simplicity this should be an all or nothing match.

Either all resources will be available in the offer matching the specified role, or not.

Also I think `"*".equals(role)` could cause true to be returned in a scenario where the resource didn't actually have the `*` role.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: Nope - the unit test unveiled that exactly this scenario failed (offer==X and requirement==X). (We only want to fail, if the offer has less resources)

snazy:
Regarding roles: so people must provide a role name? Just wanted to make sure, that if there is no role management enabled, we don't accidentally reject all offers because `*` doesn't match.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
BenWhitehead: Makes sense, you're right. We're checking the negative in this case not the positive.

BenWhitehead:
The existing code tries to read the environment variable and use the value that is provided by the user, otherwise it defaults to `*`. That means there is always a value defined and all later code has to match the specified role regardless of what it's set to.

Please remove the special null handling and just use the exact match for the value passed.

snazy:
OK, removed that special `*` role handing exceptions.
Also rebased against `rewrite`

PR #70, opened 2015-03-24
Add new config flag CASSANDRA_DATA_DIRECTORY.

BenWhitehead:
This is a temporally flag that will ultimately be replaced by persistent volumes when MESOS-1554 is released.

### DIFF
--- a/cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
+++ b/cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
@@ -14,6 +14,7 @@
 package io.mesosphere.mesos.frameworks.cassandra;
 
 import com.google.common.base.Optional;
+import com.google.common.base.Supplier;
 import io.mesosphere.mesos.frameworks.cassandra.files.FileResourceController;
 import io.mesosphere.mesos.frameworks.cassandra.util.Env;
 import io.mesosphere.mesos.util.Clock;
@@ -47,6 +48,16 @@ import static io.mesosphere.mesos.util.ProtoUtils.frameworkId;
 
 public final class Main {
     private static final Logger LOGGER = LoggerFactory.getLogger(Main.class);
+
+    private static final Supplier<String> DEFAULT_DATA_DIRECTORY = new Supplier<String>() {
+        @Override
+        public String get() {
+            LOGGER.warn("--------------------------------------------------------------------------------");
+            LOGGER.warn("| WARNING: Cassandra is configured to write data into the mesos sandbox");
+            LOGGER.warn("--------------------------------------------------------------------------------");
+            return ".";
+        }
+    };
 
     @Language("RegExp")
     private static final String userAndPass     = "[^/@]+";
@@ -92,20 +103,21 @@ public final class Main {
         final int port0 = Integer.parseInt(portOption.get());
 
         final int       executorCount           = Integer.parseInt(     Env.option("CASSANDRA_NODE_COUNT").or("3"));
-        final int       seedCount               = Integer.parseInt(     Env.option("CASSANDRA_SEED_COUNT").or("2"));
-        final double    resourceCpuCores        = Double.parseDouble(   Env.option("CASSANDRA_RESOURCE_CPU_CORES").or("2.0"));
-        final long      resourceMemoryMegabytes = Long.parseLong(       Env.option("CASSANDRA_RESOURCE_MEM_MB").or("2048"));
+        final int       seedCount               = Integer.parseInt(Env.option("CASSANDRA_SEED_COUNT").or("2"));
+        final double    resourceCpuCores        = Double.parseDouble(Env.option("CASSANDRA_RESOURCE_CPU_CORES").or("2.0"));
+        final long      resourceMemoryMegabytes = Long.parseLong(Env.option("CASSANDRA_RESOURCE_MEM_MB").or("2048"));
         final long      resourceDiskMegabytes   = Long.parseLong(       Env.option("CASSANDRA_RESOURCE_DISK_MB").or("2048"));
         final long      javaHeapMb              = Long.parseLong(       Env.option("CASSANDRA_RESOURCE_HEAP_MB").or("0"));
         final long      healthCheckIntervalSec  = Long.parseLong(       Env.option("CASSANDRA_HEALTH_CHECK_INTERVAL_SECONDS").or("60"));
         final long      bootstrapGraceTimeSec   = Long.parseLong(       Env.option("CASSANDRA_BOOTSTRAP_GRACE_TIME_SECONDS").or("120"));
         final String    cassandraVersion        =                       "2.1.2"; // TODO Env.option("CASSANDRA_VERSION").or("2.1.2");
-        final String    frameworkName           = frameworkName(        Env.option("FRAMEWORK_NAME"));
+        final String    frameworkName           = frameworkName(Env.option("FRAMEWORK_NAME"));
         final String    zkUrl                   =                       Env.option("CASSANDRA_ZK").or("zk://localhost:2181/cassandra-mesos");
-        final long      zkTimeoutMs             = Long.parseLong(       Env.option("CASSANDRA_ZK_TIMEOUT_MS").or("10000"));
+        final long      zkTimeoutMs             = Long.parseLong(Env.option("CASSANDRA_ZK_TIMEOUT_MS").or("10000"));
         final String    mesosMasterZkUrl        =                       Env.option("MESOS_ZK").or("zk://localhost:2181/mesos");
-        final long      failoverTimeout         = Long.parseLong(       Env.option("CASSANDRA_FAILOVER_TIMEOUT_SECONDS").or(String.valueOf(Period.days(7).toStandardSeconds().getSeconds())));
+        final long      failoverTimeout         = Long.parseLong(Env.option("CASSANDRA_FAILOVER_TIMEOUT_SECONDS").or(String.valueOf(Period.days(7).toStandardSeconds().getSeconds())));
         final String    mesosRole               =                       Env.option("CASSANDRA_FRAMEWORK_MESOS_ROLE").or("*");
+        final String    dataDirectory           =                       Env.option("CASSANDRA_DATA_DIRECTORY").or(DEFAULT_DATA_DIRECTORY);  // TODO: Temporary. Will be removed when MESOS-1554 is released
 
         final Matcher matcher = validateZkUrl(zkUrl);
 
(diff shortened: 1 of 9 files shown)


BenWhitehead:
@snazy please review

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
```diff
@@ -105,6 +105,7 @@ private static int _main() throws UnknownHostException {
         final String    mesosMasterZkUrl        =                       Env.option("MESOS_ZK").or("zk://localhost:2181/mesos");
         final long      failoverTimeout         = Long.parseLong(       Env.option("CASSANDRA_FAILOVER_TIMEOUT_SECONDS").or(String.valueOf(Period.days(7).toStandardSeconds().getSeconds())));
         final String    mesosRole               =                       Env.option("CASSANDRA_FRAMEWORK_MESOS_ROLE").or("*");
+        final String    dataDirectory           =                       Env.option("CASSANDRA_DATA_DIRECTORY").or("/var/lib/cassandra");  // TODO: Temporary. Will be removed when MESOS-1554 is released
```
snazy: Hm - would be better to have the env option `CASSANDRA_DATA_DIRECTORY` completely optional. Means, it should be `null` if not specified (effectively leaving the data files ephemeral, but that's ok - we can mention the fact in the readme).

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -569,6 +569,11 @@ private CassandraNode buildCassandraNode(Protos.Offer offer, boolean seed, Strin
         CassandraNode.Builder builder = CassandraNode.newBuilder()
                 .setHostname(offer.getHostname())
                 .setTargetRunState(CassandraNode.TargetRunState.RUN)
+                .addDataVolumes(
+                    DataVolume.newBuilder()
+                        .setPath(configuration.getDefaultConfigRole().getPreDefinedDataDirectory())
+                        .build()
+                )
```
snazy: Also not sure whether we should add the directory configured via the env option `CASSANDRA_DATA_DIRECTORY` to `CassandraNode.dataVolumes`. Could omit this for now as it might cause some problems when we have non-ephemeral data directories provided by mesos itself.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -212,7 +220,7 @@ message NodeLocation {
 }
 message DataVolume {
     required string path = 1;
-    required int64 sizeMb = 2;
+    optional int64 sizeMb = 2;
```
snazy: Could leave this one `required` (see `addDataVolumes` below)

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -730,6 +735,10 @@ private CassandraServerConfig buildCassandraServerConfig(
         CassandraFrameworkProtosUtils.setTaskConfig(taskConfig, configValue("rpc_port", getPortMapping(config, PORT_RPC)));
         CassandraFrameworkProtosUtils.setTaskConfig(taskConfig, configValue("seeds", SEEDS_FORMAT_JOINER.join(getSeedNodeIps())));
         CassandraFrameworkProtosUtils.setTaskConfig(taskConfig, configValue("endpoint_snitch", config.hasSnitch() ? config.getSnitch() : "GossipingPropertyFileSnitch"));
+        // data directory config
+        setTaskConfig(taskConfig, configValue("data_file_directories", newArrayList(configRole.getPreDefinedDataDirectory() + "/data")));
+        setTaskConfig(taskConfig, configValue("commitlog_directory", configRole.getPreDefinedDataDirectory() + "/commitlog"));
+        setTaskConfig(taskConfig, configValue("saved_caches_directory", configRole.getPreDefinedDataDirectory() + "/saved_caches"));
```
snazy: Just two nits:
* would like to have a `TODO` comment to remove that once Mesos has data-volumes support
* `setTaskConfig` and `CassandraFrameworkProtosUtils.setTaskConfig` (two styles). Nicer to use only one variant (no preference to either).

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -88,6 +88,14 @@ message CassandraConfigRole {
...
     // mesos role to be used to receive resource offers
     optional string mesosRole = 12;
+
+    /**
+     * A pre-defined data directory specifying where cassandra should write it's data.
+     * NOTE:
+     * This field will be removed once MESOS-1554 is released and the framework will
+     * be able to allocate the data volume itself.
+     */
+    optional string preDefinedDataDirectory = 13;
```
snazy: Can we remove a field from protobuf? Maybe we just don't use it once MESOS-1554 is in a release.

marathon.json
```diff
@@ -18,7 +18,8 @@
     "CASSANDRA_RESOURCE_MEM_MB": "2048",
     "CASSANDRA_RESOURCE_DISK_MB": "2048",
     "CASSANDRA_HEALTH_CHECK_INTERVAL_SECONDS": "60",
-    "CASSANDRA_ZK_TIMEOUT_MS": "10000"
+    "CASSANDRA_ZK_TIMEOUT_MS": "10000",
+    "CASSANDRA_DATA_DIRECTORY": "/var/lib/cassandra"
```
snazy: Can we make this optional so that a user can configure it if needed?

snazy:
Generally ok, but I'd prefer to make the configuration of `CASSANDRA_DATA_DIRECTORY` explicit - i.e. leave it up to the owner whether the nodes should be _screwed up_ with a directory. So leaving the directories ephemeral by default.

air:
From https://wiki.apache.org/cassandra/GettingStarted

> Since Cassandra 2.1, the tar.gz download has shipped with the log and data directories defaulting to the Cassandra directory. Versions prior defaulted to /var/log/cassandra and /var/lib/cassandra/.

BenWhitehead:
Okay I'll update the code so that the default is to leave the data in the sandbox.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
BenWhitehead: This needs to be optional since we're not able to constrain or verify the size of the data volume since we're not using persistent volumes.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
BenWhitehead: I think we need to keep track of it now because there is the potential that we will need to support migrating the data from a previously defined directory to a persistent volume.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
BenWhitehead: A field can be removed from a protobuf later on, we just have to make sure that we don't re-use `13` as a fields cardinal value.  I'll change this to be something much higher and unlikely to ever be hit naturally.

BenWhitehead:
@snazy updated based on feedback. One last review then I'll rebase and squash before merging.

snazy:
Yes, `/var/lib/cassandra/` is the standard for pre-2.1 tarballs and all current packaged (Debian) installations. The tarball has been changed in 2.1 for dev and test convenience.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
snazy: Ah - sure. My bad.

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
snazy: OK - then we have the optional {{sizeMb}} field to check whether it needs to be migrated.

cassandra-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/Main.java
```diff
@@ -48,6 +49,16 @@
 public final class Main {
     private static final Logger LOGGER = LoggerFactory.getLogger(Main.class);
 
+    private static final Supplier<String> DEFAULT_DATA_DIRECTORY = new Supplier<String>() {
+        @Override
+        public String get() {
+            LOGGER.warn("--------------------------------------------------------------------------------");
+            LOGGER.warn("| WARNING: Cassandra is configured to write data into the mesos sandbox");
+            LOGGER.warn("--------------------------------------------------------------------------------");
```
snazy: +1 on this :)

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -88,6 +88,14 @@ message CassandraConfigRole {
...
     // mesos role to be used to receive resource offers
     optional string mesosRole = 12;
+
+    /**
+     * A pre-defined data directory specifying where cassandra should write it's data.
+     * NOTE:
+     * This field will be removed once MESOS-1554 is released and the framework will
+     * be able to allocate the data volume itself.
+     */
+    optional string preDefinedDataDirectory = 999999;
```
snazy: I'd also be fine with `13`. We will run into similar issues with fields we are not yet aware of. IMO we only need a "marker" comment that says: "Hey, start with 14".

cassandra-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/CassandraCluster.java
```diff
@@ -569,6 +569,11 @@ private CassandraNode buildCassandraNode(Protos.Offer offer, boolean seed, Strin
         CassandraNode.Builder builder = CassandraNode.newBuilder()
                 .setHostname(offer.getHostname())
                 .setTargetRunState(CassandraNode.TargetRunState.RUN)
+                .addDataVolumes(
+                    DataVolume.newBuilder()
+                        .setPath(configuration.getDefaultConfigRole().getPreDefinedDataDirectory())
+                        .build()
```
snazy: NIT: can remove the last `.build()`

snazy:
Just some minor then ready for merge.

BenWhitehead:
Fixed based on feedback. Rebase squash pre-merge.

PR #72, opened 2015-03-26
Cleanups

snazy:
Some more cleanups separated in different commits:

* Unit tests run on localhost (NIT - feel free to omit this)
* some NIT commits (ed4dd60, e936899)
* Add check of resources in offer for new nodes (don't acquire nodes that can never fulfil resource requirements)
* Make `CassandraNode.cassandraNodeExecutor` required and initialize it when building `CassandraNode` (we never clear that field)
* bugfix to only issue node-jobs to nodes that have target-run-state `RUN` and are considered live
* fix slow shutdown own server task (close of JMX connection takes very long if remote is gone)
* fix handling in scheduler when mesos-slave is killed

### DIFF
--- a/cassandra-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/ApiControllerTest.java
+++ b/cassandra-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/ApiControllerTest.java
@@ -696,7 +696,7 @@ public class ApiControllerTest extends AbstractCassandraSchedulerTest {
 
         try {
             try (ServerSocket sock = new ServerSocket(0)) {
-                httpServerBaseUri = URI.create(String.format("http://%s:%d/", formatInetAddress(InetAddress.getLocalHost()), sock.getLocalPort()));
+                httpServerBaseUri = URI.create(String.format("http://%s:%d/", formatInetAddress(InetAddress.getLoopbackAddress()), sock.getLocalPort()));
             }
 
             final ResourceConfig rc = new ResourceConfig()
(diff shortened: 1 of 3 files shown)


cassandra-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/ApiControllerTest.java
```diff
@@ -696,7 +696,7 @@ public void cleanState() {
 
         try {
             try (ServerSocket sock = new ServerSocket(0)) {
-                httpServerBaseUri = URI.create(String.format("http://%s:%d/", formatInetAddress(InetAddress.getLocalHost()), sock.getLocalPort()));
+                httpServerBaseUri = URI.create(String.format("http://%s:%d/", formatInetAddress(InetAddress.getLoopbackAddress()), sock.getLocalPort()));
```
BenWhitehead: Why is this necessary? I've not has issues running tests in local mode, nor have I seen issues when running via teamcity.

BenWhitehead:
810e386, 9c0c3f3 and 975110b seem okay to merge, you can reset to 975110b in this PR after you get all the other changes moved to their respective PRs.

Right now I'm inclined to defer 00319db until a later version. the scenario that is attempted to be avoided is still possible until such time as dynamic reservations is in mesos.  The operator will have to absorb some of the responsibility until the framework can do it on it's own.  There is also the option that an operator could create a resource reservation for the role cassandra runs as which can be used to ensure it has the expected resources.

171bbd7 should be removed

Please break ff0a972 into it's own pr and provide a little more context around what is being fixed.

Please break 51be049 into it's own PR.

Please move 5cc25cd into it's own PR.

Remove commit 371c17d it is being tracked outside the code base.

Please open a new PR with 959ad0d and 1c9433d in it.

None of these changes really look like they depend on each other so please have each PR be based from current `rewrite:HEAD`.

cassandra-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/ApiControllerTest.java
snazy: It's usually not necessary. I did it while traveling without any network (so OSX refused to resolved the host name)

snazy:
OK - branch is reset and some more PRs created

PR #85, opened 2015-04-02
split ApiController

snazy:
PR splits ApiController (and ApiControllerTest) into separate classes (in a separate `api` package).
No functional changes.

PR #86, opened 2015-04-02
Streaming api responses

snazy:
Let all REST API responses stream instead of building it in a possibly large string buffer (StringWriter).

Nice side effect: the code looks much cleaner

Based on PR #85


BenWhitehead:
Looks ready to merge after a rebase

snazy:
rebased

PR #87, opened 2015-04-02
add NEWS.txt

snazy:
Just adds 
* NEWS.txt to track new functionality and breaking changes (things, that people need to know)
* CHANGES.txt as a change-log


BenWhitehead:
We're going to leverage githubs change log/tag feature for this functionality rather than having it in the repo itself.

PR #98, opened 2015-04-24
Node scaling

BenWhitehead:


BenWhitehead:
Tested scale up on 100 node cluster.

Framework was successfully able to launch all 100 nodes after ~3.5 hrs of time.

Rollout time is approximately 3.5 hrs due to configured timeout of 120 seconds between each node.

100 * 120 = 12,000s = 200m = 3.33h

PR #100, opened 2015-04-28
Only repair primary range

snazy:
It's only necessary to repair a node's primary range (which will 'propagate' to the replica nodes). That's sufficient since the cluster-job does the repair on all nodes.

Additionally only repair in the 'local DC' (currently irrelevant for C* on Mesos).


BenWhitehead:
Thanks @snazy

PR #103, opened 2015-05-13
Update link to REST API docs

pitluga:
The current link was broken. This seems like where it should go.


BenWhitehead:
thanks @pitluga

PR #109, opened 2015-06-01
Update to cassandra version 2.1.5

BenWhitehead:


BenWhitehead:
@air Tested on a dcos cluster and a custom universe definition. Is this something we want to merge now or hold off?

air:
As discussed, hold for now - we can't get enough testing in before code freeze.

BenWhitehead:
A newer version is what will be updated to when it happens.

PR #114, opened 2015-07-03
mesos_user_http_host

dmitrypekar:
Hey, 

This is first PR. It is required for two purposes:
- dev (need possibility to specify http host name);
- dcos compat (need possibility to specify framework user);

I've added support of following options:
- mesos_user - specifies framework user (defaults to "");
- host - specifies public address of http host (defaults to "localhost");

Both those options are required if running Scheduler outside of mesos hosts 
(for instance when using vagrant for mesos and running Scheduler inside IDE).

Please review & merge.


BenWhitehead:
Thanks @dmitrypekar

PR #120, opened 2015-07-13
multi-dc-seed-sync

dmitrypekar:
Hey, Ben.

Just pushed final changes to support dc syncing.
For now I am going to run some more tests (99% that they would pass).
During that time you can review the code and post your comments.

### DIFF
--- a/cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
+++ b/cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
@@ -45,6 +45,8 @@ import org.slf4j.bridge.SLF4JBridgeHandler;
 
 import java.net.URI;
 import java.net.UnknownHostException;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Handler;
 import java.util.logging.Level;
@@ -52,6 +54,8 @@ import java.util.logging.LogManager;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static com.google.common.collect.Lists.newArrayList;
+import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.ExternalDc;
 import static io.mesosphere.mesos.util.ProtoUtils.frameworkId;
 
 public final class Main {
@@ -133,6 +137,7 @@ public final class Main {
         final String    defaultRack                 =                       Env.option("CASSANDRA_DEFAULT_RACK").or("RACK0");
         final String    defaultDc                   =                       Env.option("CASSANDRA_DEFAULT_DC").or("DC0");
 
+        final List<ExternalDc> externalDcs = getExternalDcs(Env.filterStartsWith("CASSANDRA_EXTERNAL_DC_", true));
         final Matcher matcher = validateZkUrl(zkUrl);
 
         final State state = new ZooKeeperState(
@@ -163,7 +168,8 @@ public final class Main {
             jmxLocal,
             jmxNoAuthentication,
             defaultRack,
-            defaultDc);
+            defaultDc,
+            externalDcs);
 
 
         final FrameworkInfo.Builder frameworkBuilder =
@@ -184,6 +190,7 @@ public final class Main {
         final Clock clock = new SystemClock();
         final PersistedCassandraClusterHealthCheckHistory healthCheckHistory = new PersistedCassandraClusterHealthCheckHistory(state);
         final PersistedCassandraClusterState clusterState = new PersistedCassandraClusterState(state);
+        final SeedManager seedManager = new SeedManager(configuration, new ObjectMapper(), new SystemClock());
         final CassandraCluster cassandraCluster = new CassandraCluster(
             clock,
             httpServerBaseUri.toString(),
@@ -191,7 +198,8 @@ public final class Main {
             clusterState,
             healthCheckHistory,
             new PersistedCassandraClusterJobs(state),
-            configuration
+            configuration,
+            seedManager
         );
         final HealthReportService healthReportService = new HealthReportService(
             clusterState,
@@ -201,7 +209,8 @@ public final class Main {
         );
         final Scheduler scheduler = new CassandraScheduler(
             configuration,
-            cassandraCluster
+            cassandraCluster,
+            clock
         );
 
         final JsonFactory factory = new JsonFactory();
@@ -239,6 +248,8 @@ public final class Main {
             driver = new MesosSchedulerDriver(scheduler, frameworkBuilder.build(), mesosMasterZkUrl);
         }
 
+        seedManager.startSyncingSeeds(60);
+
         final int status;
         switch (driver.run()) {
             case DRIVER_STOPPED:
@@ -259,6 +270,20 @@ public final class Main {
         // Ensure that the driver process terminates.
         driver.stop(true);
         return status;
+    }
+
+    static List<ExternalDc> getExternalDcs(Map<String, String> dcOpts) {
+        final List<ExternalDc> externalDcs = newArrayList();
+
+        for (final String key: dcOpts.keySet()) {
+            externalDcs.add(
+                ExternalDc.newBuilder()
+                    .setName(key)
+                    .setUrl(dcOpts.get(key))
+                    .build()
+            );
+        }
+        return externalDcs;
     }
 
     static Matcher validateZkUrl(final String zkUrl) {
--- a/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
+++ b/cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
@@ -67,6 +67,26 @@ message CassandraFrameworkConfiguration {
      * target number of seed nodes.
      */
     optional int32 targetNumberOfSeeds = 9;
+
+    /** List of external DCs running other rings of this cluster */
+    repeated ExternalDc externalDcs = 10;
+}
+
+/**
+ * External DC, running other ring of the same cluster
+ */
+message ExternalDc {
+    /** DC name */
+    optional string name = 1;
+
+    /** REST API url */
+    optional string url = 2;
+
+    /** Time of last seed fetching */
+    optional int64 seedFetchTime = 3;
+
+    /** List of fetched seeds */
+    repeated string seeds = 4;
 }
 
 /**
--- a/cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
+++ b/cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
@@ -114,6 +114,8 @@ public final class CassandraCluster {
     private final PersistedCassandraFrameworkConfiguration configuration;
     @NotNull
     private final PersistedCassandraClusterJobs jobsState;
+    @NotNull
+    private final SeedManager seedManager;
 
     @NotNull
     private final Map<ClusterJobType, ClusterJobHandler> clusterJobHandlers;
@@ -138,7 +140,8 @@ public final class CassandraCluster {
         @NotNull final PersistedCassandraClusterState clusterState,
         @NotNull final PersistedCassandraClusterHealthCheckHistory healthCheckHistory,
         @NotNull final PersistedCassandraClusterJobs jobsState,
-        @NotNull final PersistedCassandraFrameworkConfiguration configuration
+        @NotNull final PersistedCassandraFrameworkConfiguration configuration,
+        @NotNull final SeedManager seedManager
     ) {
         this.clock = clock;
         this.httpServerBaseUrl = httpServerBaseUrl;
@@ -147,6 +150,7 @@ public final class CassandraCluster {
         this.healthCheckHistory = healthCheckHistory;
         this.jobsState = jobsState;
         this.configuration = configuration;
+        this.seedManager = seedManager;
 
         clusterJobHandlers = new EnumMap<>(ClusterJobType.class);
         clusterJobHandlers.put(ClusterJobType.CLEANUP, new NodeTaskClusterJobHandler(this, jobsState));
@@ -162,6 +166,11 @@ public final class CassandraCluster {
     @NotNull
     public PersistedCassandraFrameworkConfiguration getConfiguration() {
         return configuration;
+    }
+
+    @NotNull
+    public SeedManager getSeedManager() {
+        return seedManager;
     }
 
     @Nullable
@@ -316,8 +325,12 @@ public final class CassandraCluster {
     }
 
     @NotNull
-    public List<String> getSeedNodeIps() {
-        return CassandraFrameworkProtosUtils.getSeedNodeIps(clusterState.nodes());
+    public List<String> getSeedNodeIps(final boolean addExternal) {
+        final List<String> seeds = CassandraFrameworkProtosUtils.getSeedNodeIps(clusterState.nodes());
+        if (addExternal) {
+            seeds.addAll(seedManager.getSeeds());
+        }
+        return seeds;
     }
 
     @NotNull
@@ -561,7 +574,7 @@ public final class CassandraCluster {
         CassandraFrameworkProtosUtils.setTaskConfig(taskConfig, configValue("ssl_storage_port", getPortMapping(config, PORT_STORAGE_SSL)));
         CassandraFrameworkProtosUtils.setTaskConfig(taskConfig, configValue("native_transport_port", getPortMapping(config, PORT_NATIVE)));
         CassandraFrameworkProtosUtils.setTaskConfig(taskConfig, configValue("rpc_port", getPortMapping(config, PORT_RPC)));
-        CassandraFrameworkProtosUtils.setTaskConfig(taskConfig, configValue("seeds", SEEDS_FORMAT_JOINER.join(getSeedNodeIps())));
+        CassandraFrameworkProtosUtils.setTaskConfig(taskConfig, configValue("seeds", SEEDS_FORMAT_JOINER.join(getSeedNodeIps(true))));
         CassandraFrameworkProtosUtils.setTaskConfig(taskConfig, configValue("endpoint_snitch", config.hasSnitch() ? config.getSnitch() : "GossipingPropertyFileSnitch"));
         // data directory config
         // TODO: Update the logic here for defining data directories when mesos persistent volumes are released
--- a/cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraScheduler.java
+++ b/cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraScheduler.java
@@ -22,6 +22,7 @@ import com.google.common.base.Optional;
 import com.google.common.collect.FluentIterable;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
+import io.mesosphere.mesos.util.Clock;
 import org.apache.mesos.Protos.*;
 import org.apache.mesos.Scheduler;
 import org.apache.mesos.SchedulerDriver;
@@ -55,13 +56,17 @@ public final class CassandraScheduler implements Scheduler {
     private final PersistedCassandraFrameworkConfiguration configuration;
     @NotNull
     private final CassandraCluster cassandraCluster;
+    @NotNull
+    private final Clock clock;
 
     public CassandraScheduler(
         @NotNull final PersistedCassandraFrameworkConfiguration configuration,
-        @NotNull final CassandraCluster cassandraCluster
+        @NotNull final CassandraCluster cassandraCluster,
+        @NotNull final Clock clock
     ) {
         this.configuration = configuration;
         this.cassandraCluster = cassandraCluster;
+        this.clock = clock;
     }
 
     @Override
@@ -88,6 +93,7 @@ public final class CassandraScheduler implements Scheduler {
         if (LOGGER.isDebugEnabled()) {
             LOGGER.debug("> resourceOffers(driver : {}, offers : {})", driver, protoToString(offers));
         }
+
         for (final Offer offer : offers) {
             final Marker marker = MarkerFactory.getMarker("offerId:" + offer.getId().getValue() + ",hostname:" + offer.getHostname());
             final boolean offerUsed = evaluateOffer(driver, marker, offer);
--- a/cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/PersistedCassandraFrameworkConfiguration.java
+++ b/cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/PersistedCassandraFrameworkConfiguration.java
@@ -26,8 +26,11 @@ import org.apache.mesos.state.State;
 import org.jetbrains.annotations.NotNull;
 import org.joda.time.Duration;
 
-public final class PersistedCassandraFrameworkConfiguration extends StatePersistedObject<CassandraFrameworkConfiguration> {
+import java.util.List;
 
+import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.ExternalDc;
+
+public final class PersistedCassandraFrameworkConfiguration extends StatePersistedObject<CassandraFrameworkConfiguration> {
     public PersistedCassandraFrameworkConfiguration(
         @NotNull final State state,
         @NotNull final String frameworkName,
@@ -45,7 +48,8 @@ public final class PersistedCassandraFrameworkConfiguration extends StatePersist
         final boolean jmxLocal,
         final boolean jmxNoAuthentication,
         @NotNull final String defaultRack,
-        @NotNull final String defaultDc
+        @NotNull final String defaultDc,
+        @NotNull final List<ExternalDc> externalDcs
     ) {
         super(
             "CassandraFrameworkConfiguration",
@@ -79,6 +83,7 @@ public final class PersistedCassandraFrameworkConfiguration extends StatePersist
                         .setBootstrapGraceTimeSeconds(bootstrapGraceTimeSec)
                         .setTargetNumberOfNodes(executorCount)
                         .setTargetNumberOfSeeds(seedCount)
+                        .addAllExternalDcs(externalDcs)
                         .build();
                 }
             },
--- a/cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
+++ b/cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
@@ -1,0 +1,152 @@
+package io.mesosphere.mesos.frameworks.cassandra.scheduler;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Joiner;
+import io.mesosphere.mesos.util.Clock;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.CassandraFrameworkConfiguration;
+import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.ExternalDc;
+
+public class SeedManager {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SeedManager.class);
+
+    @NotNull
+    private final PersistedCassandraFrameworkConfiguration configuration;
+
+    @NotNull
+    private final ObjectMapper objectMapper;
+
+    @NotNull
+    private final ScheduledExecutorService scheduledExecutorService;
+    private ScheduledFuture<?> syncingTask;
+
+    @NotNull
+    private final Clock clock;
+
+    public SeedManager(@NotNull final PersistedCassandraFrameworkConfiguration configuration,
+                       @NotNull final ObjectMapper objectMapper,
+                       @NotNull final Clock clock) {
+        this.configuration = configuration;
+        this.objectMapper = objectMapper;
+        this.clock = clock;
+        this.scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+    }
+
+    @NotNull
+    public List<String> getSeeds() {
+        final List<String> seeds = newArrayList();
+
+        for (final ExternalDc dc : configuration.get().getExternalDcsList()) {
+            seeds.addAll(dc.getSeedsList());
+        }
+
+        return seeds;
+    }
+
+    public void startSyncingSeeds(final long periodSeconds) {
+        stopSyncingSeeds();
+        LOGGER.debug("Scheduling background syncing task to run every {} seconds", periodSeconds);
+        syncingTask = scheduledExecutorService.scheduleAtFixedRate(
+                new SyncingTask(),
+                0,
+                periodSeconds,
+                TimeUnit.SECONDS
+        );
+    }
+
+    private void stopSyncingSeeds() {
+        if (syncingTask != null) {
+            LOGGER.debug("Stopping scheduled background syncing task");
+            syncingTask.cancel(true);
+            syncingTask = null;
+        }
+    }
+
+    public void syncSeeds() {
+        LOGGER.info("Syncing seeds ...");
+        for (ExternalDc dc : configuration.get().getExternalDcsList()) {
+            fetchSeeds(dc);
+        }
+    }
+
+    private void fetchSeeds(@NotNull final ExternalDc dc) {
+        String url = dc.getUrl();
+        if (url.endsWith("/")) url = url.substring(0, url.length() - 1);
+        url += "/node/seed/all";
+
+        JsonNode node = fetchJson(url);
+        if (node == null) return;
+
+        JsonNode seedsArr = node.get("seeds");
+        List<String> seeds = newArrayList();
+        for (JsonNode ip : seedsArr) {
+            seeds.add(ip.asText());
+        }
+
+        LOGGER.info("Fetched seeds for dc \"" + dc.getName() + "\": " + Joiner.on(", ").join(seeds));
+        updateDcSeeds(dc, seeds);
+    }
+
+    @Nullable
+    protected JsonNode fetchJson(@NotNull final String url) {
+        HttpURLConnection c = null;
+        try {
+            c = (HttpURLConnection)new URL(url).openConnection();
+            return objectMapper.readTree(c.getInputStream());
+        } catch (IOException e) {
+            LOGGER.warn("", e);
+        } finally {
+            if (c != null) c.disconnect();
+        }
+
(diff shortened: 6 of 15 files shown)


cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
```diff
@@ -36,6 +40,17 @@ public static String get(@NotNull final String key) {
     }
 
     @NotNull
+    public static Map<String, String> optionsStart(@NotNull final String prefix, final boolean trimPrefix) {
```
BenWhitehead: please rename method to `filterStartsWith(@NotNull final String prefix, final boolean trimPrefix)`.

The name option on line 38 was used to represent an optional value for a given key, but in this scenario we're already returning a 0-many data structure we won't need option in the name.  In actuality we're really filtering all of the environment down to those entries whose key begins with `prefix`.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
```diff
@@ -36,6 +40,17 @@ public static String get(@NotNull final String key) {
...
+        final Map<String, String> options = newHashMap();
+
+        for (final String key : System.getenv().keySet())
```
BenWhitehead: Add curly braces to `for` and `if` below.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
```diff
@@ -36,6 +40,17 @@ public static String get(@NotNull final String key) {
     }
 
     @NotNull
+    public static Map<String, String> optionsStart(@NotNull final String prefix, final boolean trimPrefix) {
+        final Map<String, String> options = newHashMap();
```
BenWhitehead: rename variable to something more appropriate of "things that haven't been filtered out".

cassandra-mesos-scheduler/pom.xml
```diff
@@ -80,6 +80,11 @@
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-guava</artifactId>
         </dependency>
+        <dependency>
```
BenWhitehead: Jackson is already in the code base, please use that rather than adding another json library.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
```diff
@@ -1320,4 +1326,36 @@ static long secondsUntilNextPossibleServerLaunch(final long now, final long next
...
+            JSONObject obj = (JSONObject) new JSONParser().parse(new InputStreamReader(c.getInputStream(), "utf-8"));
+
+            JSONArray seedsArr = (JSONArray) obj.get("seeds");
+            for (Object ip : seedsArr) seeds.add("" + ip);
+        } catch (IOException | ParseException e) {
+            LOGGER.warn("", e);
+        } finally {
+            if (c != null) c.disconnect();
+        }
+    }
```
BenWhitehead: Please move `syncSeedsIfRequired` and `fetchSeeds` into another service class (perhaps `SeedManager` or something). Then pass that new service class into `CassandraCluster` as a constructor argument.

In the new class you can pass in the jackson ObjectMapper to use for json deserialization to a simple struct.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
```diff
@@ -1320,4 +1326,36 @@ static long secondsUntilNextPossibleServerLaunch(final long now, final long next
         }
         return millisUntilNext / 1000;
     }
+
+    public void syncSeedsIfRequired() {
+        long lastTime = clusterState.get().getLastSeedFetchTime();
+        long now = System.currentTimeMillis();
```
BenWhitehead: Please use `Clock` instead of `System.currentTimeMillis()` so that this code is more easily tested.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
```diff
@@ -1320,4 +1326,36 @@ static long secondsUntilNextPossibleServerLaunch(final long now, final long next
...
+        if (now - lastTime < TimeUnit.MINUTES.toMillis(1)) return;
+
+        LOGGER.info("Syncing seeds from external DCs ...");
+        List<String> seeds = newArrayList();
+        for (ExternalDc dc : configuration.get().getExternalDcsList())
```
BenWhitehead: Add curly braces to `for`

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
```diff
@@ -1320,4 +1326,36 @@ static long secondsUntilNextPossibleServerLaunch(final long now, final long next
...
+    }
+
+    private void fetchSeeds(String url, List<String> seeds) {
+        HttpURLConnection c = null;
+        try {
+            if (url.endsWith("/")) url = url.substring(0, url.length() - 1);
+            url += "/node/seed/all";
+
+            c = (HttpURLConnection)new URL(url).openConnection();
+            JSONObject obj = (JSONObject) new JSONParser().parse(new InputStreamReader(c.getInputStream(), "utf-8"));
```
BenWhitehead: replace JSONParser with Jackson ObjectMapper and simple struct.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
```diff
@@ -1320,4 +1326,36 @@ static long secondsUntilNextPossibleServerLaunch(final long now, final long next
...
+
+            JSONArray seedsArr = (JSONArray) obj.get("seeds");
+            for (Object ip : seedsArr) seeds.add("" + ip);
```
BenWhitehead: add curly braces to `for` and move statement to it's own line.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
```diff
@@ -187,6 +201,13 @@ message CassandraClusterState {
      * List of IP addresses of nodes that need to be replaced.
      */
     repeated string replaceNodeIps = 4;
+
+    /** Time of last fetching of seeds from ExternalDcs */
+    optional int64 lastSeedFetchTime = 5;
+
+    /** List of external seeds fetched from ExternalDc(s) via REST API */
+    repeated string externalSeeds = 6;
```
BenWhitehead: Can we move these properties into the new `ExternalDc` message so that they can be tracked for each datacenter individually.

I'm imagining a scenario, where you are running cassandra in 3 DC and one of the DC becomes unavailable due to some network outage.  We should probably continue use the last known seed set for that DC even if we aren't able to re-query it at this point in time.

This could also be something we add to the health check report to surface whether this cluster can communicate with the other clusters and when the last communication was.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
```diff
@@ -316,8 +320,10 @@ public void recordHealthCheck(@NotNull final String executorId, @NotNull final H
     }
 
     @NotNull
-    public List<String> getSeedNodeIps() {
-        return CassandraFrameworkProtosUtils.getSeedNodeIps(clusterState.nodes());
+    public List<String> getSeedNodeIps(final boolean addExternal) {
+        final List<String> seeds = CassandraFrameworkProtosUtils.getSeedNodeIps(clusterState.nodes());
+        if (addExternal) seeds.addAll(clusterState.getExternalSeeds());
```
BenWhitehead: Add curly braces to `if` and move the statement to it's own line.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
```diff
@@ -36,6 +40,17 @@ public static String get(@NotNull final String key) {
...
+
+        for (final String key : System.getenv().keySet())
+            if (key.startsWith(prefix))
+                options.put(trimPrefix ? key.substring(prefix.length()) : key, System.getenv(key));
```
BenWhitehead: Rather than looking up the value from the system again, change your iteration to be `entrySet` then just use `entry.getValue`

cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
```diff
@@ -133,6 +137,17 @@ private static int _main() throws UnknownHostException {
...
+        final Map<String, String> dcOpts = Env.optionsStart("CASSANDRA_EXTERNAL_DC_", true);
+        final List<ExternalDc> externalDcs = newArrayList();
+        for (final String key: dcOpts.keySet()) {
+            externalDcs.add(
+                ExternalDc.newBuilder()
+                    .setName(key)
+                    .setUrl(dcOpts.get(key))
+                    .build()
+            );
+        }
```
BenWhitehead: Please move lines 141:149 to their own (package private) method and write a unit test for this.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
```diff
@@ -36,6 +40,17 @@ public static String get(@NotNull final String key) {
...
+
+        return options;
```
BenWhitehead: Please add some unit tests for this method (you may need to refactor the signature to pass in a map rather than it just using the system env.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
BenWhitehead: Please add unit tests for the new class when you create it.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
dmitrypekar: fixed

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
dmitrypekar: fixed

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
dmitrypekar: fixed

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
dmitrypekar: fixed

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
dmitrypekar: fixed

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
dmitrypekar: fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
dmitrypekar: fixed

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
dmitrypekar: fixed

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
dmitrypekar: fixed

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
dmitrypekar: fixed

cassandra-mesos-scheduler/pom.xml
dmitrypekar: fixed

cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
dmitrypekar: fixed

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/util/Env.java
dmitrypekar: IMHO, that test would not add a lot of value:
- the method is simple;
- the method has low probability of regression;

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
dmitrypekar: fixed

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/PersistedCassandraFrameworkConfiguration.java
```diff
@@ -26,7 +26,12 @@
 import org.jetbrains.annotations.NotNull;
 import org.joda.time.Duration;
 
+import java.util.List;
+
+import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.ExternalDc;
+
 public final class PersistedCassandraFrameworkConfiguration extends StatePersistedObject<CassandraFrameworkConfiguration> {
+    @NotNull private final SeedManager seedManager;
```
BenWhitehead: This class should not have a service dependency other than `State`.

This field should be passed to `CassandraCluster` as a constructor argument that can then be set to a field.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/PersistedCassandraFrameworkConfiguration.java
```diff
@@ -99,6 +106,13 @@ public CassandraFrameworkConfiguration apply(final byte[] input) {
                 }
             }
         );
+
+        seedManager = new SeedManager(this);
```
BenWhitehead: Move construction to `Main` and pass it down from there.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
```diff
@@ -0,0 +1,110 @@
...
+import java.net.URL;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.CassandraFrameworkConfiguration;
+import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.ExternalDc;
+
+public class SeedManager {
+    public static final long FETCH_PERIOD = TimeUnit.MINUTES.toMillis(1);
```
BenWhitehead: Please make package private.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(SeedManager.class);
+
+    @NotNull
+    private final PersistedCassandraFrameworkConfiguration configuration;
+
+    public SeedManager(@NotNull PersistedCassandraFrameworkConfiguration configuration) {
```
BenWhitehead: The `Clock` should be passed into this constructor and then be set to a field (this can easily be done when this is constructed in `Main`.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
```diff
@@ -0,0 +1,110 @@
...
+        final List<String> seeds = newArrayList();
+
+        for (final ExternalDc dc : configuration.get().getExternalDcsList()) {
+            seeds.addAll(dc.getSeedsList());
+        }
+
+        return seeds;
+    }
+
+    public void syncSeeds(long now) {
```
BenWhitehead: Remove this parameter, it won't be needed when `Clock` is passed to this class.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
```diff
@@ -0,0 +1,110 @@
...
+        for (ExternalDc dc : configuration.get().getExternalDcsList()) {
+            fetchSeedsIfRequired(dc, now);
+        }
+    }
+
+    private void fetchSeedsIfRequired(ExternalDc dc, long now) {
```
BenWhitehead: Remove `now` parameter, and replace with `Clock` once it's passed into this class.

BenWhitehead: please add `@NotNull final` to `dc` parameter.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
```diff
@@ -0,0 +1,110 @@
...
+        List<String> seeds = newArrayList();
+        for (JsonNode ip : seedsArr) {
+            seeds.add(ip.asText());
+        }
+
+        LOGGER.info("Got seeds for dc \"" + dc.getName() + "\": " + Joiner.on(", ").join(seeds));
+        updateDcSeeds(dc, seeds, now);
+    }
+
+    protected JsonNode fetchJson(String url) {
```
BenWhitehead: Please add `@NotNull final` to `url` parameter.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
```diff
@@ -0,0 +1,110 @@
...
+        HttpURLConnection c = null;
+        try {
+            c = (HttpURLConnection)new URL(url).openConnection();
+            ObjectMapper mapper = new ObjectMapper();
```
BenWhitehead: `ObjectMapper` should be passed into the constructor of this class. (Constructing `ObjectMapper`s is quite costly in jackson and should only be done once)

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
BenWhitehead: Please add `@Nullable` annotation to this method.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
```diff
@@ -0,0 +1,110 @@
...
+        } catch (IOException e) {
+            LOGGER.warn("", e);
+        } finally {
+            if (c != null) c.disconnect();
+        }
+
+        return null;
+    }
+
+    private int externalDxIdx(final String name) {
```
BenWhitehead: Typo: `externalDxIdx` instead of `externalDcIdx`  (`x` vs `c` in Dc)

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
```diff
@@ -0,0 +1,110 @@
...
+            if (dc.getName().equals(name)) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private void updateDcSeeds(final ExternalDc dc, final List<String> seeds, final long time) {
+        int idx = externalDxIdx(dc.getName());
+        if (idx == -1) throw new IllegalArgumentException("Unknown dc " + dc.getName());
```
BenWhitehead: Add curly braces to `if` statement and move statement to it's own line.

Also, is there some other way we can deal with this other than throwing an exception? This code is confusing to me, can you explain what it's doing?

cassandra-mesos-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManagerTest.java
```diff
@@ -0,0 +1,91 @@
...
+                "RACK1",
+                "DC1",
+                Arrays.asList(ExternalDc.newBuilder()
+                        .setName("dc")
+                        .setUrl("http://dc")
+                        .setSeedFetchTime(-SeedManager.FETCH_PERIOD)
+                        .build())
+        );
+
+        seedManager = new SeedManager(config) {
```
BenWhitehead: Can the original code be refactors so that this inline overriding doesn't have to take place?  This override shadows a method that is actually pretty important.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
```diff
@@ -0,0 +1,110 @@
...
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.CassandraFrameworkConfiguration;
+import static io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.ExternalDc;
+
+public class SeedManager {
```
BenWhitehead: Please make this class final.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraScheduler.java
```diff
@@ -88,6 +93,9 @@ public void resourceOffers(final SchedulerDriver driver, final List<Offer> offer
         if (LOGGER.isDebugEnabled()) {
             LOGGER.debug("> resourceOffers(driver : {}, offers : {})", driver, protoToString(offers));
         }
+
+        configuration.getSeedManager().syncSeeds(clock.now().getMillis());
```
BenWhitehead: I don't think this should be called on every resource offer.  Instead `SeedManager` should be updated to have it's constructor receive a `java.util.concurrent.ScheduledExecutorService` and to start a background task to call `syncSeeds` on a periodic basis.

And example elsewhere in the code that is following this same pattern can be found in [`CassandraExecutor#startCheckingHealth`](https://github.com/mesosphere/cassandra-mesos/blob/master/cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/CassandraExecutor.java#L514-L519)

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/PersistedCassandraFrameworkConfiguration.java
dmitrypekar: Fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/PersistedCassandraFrameworkConfiguration.java
dmitrypekar: Fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Note sure about this. Avg creation of ObjectMapper takes ~ 0.031 ms on my laptop. Please see the output:
```
import com.fasterxml.jackson.databind.ObjectMapper;

public class T {
    public static void main(String[] args) {
        long codeRef = 0, start = System.currentTimeMillis();
        int creations = 10000;

        for (int i = 0; i < creations; i++) {
            ObjectMapper mapper = new ObjectMapper();
            codeRef += mapper.hashCode();
        }

        long time = System.currentTimeMillis() - start;
        double avg = Math.floor(time / (double)creations * 1000) / 1000;

        System.out.println(codeRef);
        System.out.println(creations + " creations of ObjectMapper took " + time + " ms, avg took " + avg + " ms");
    }
}
```
Output:
```
10700061484741
10000 creations of ObjectMapper took 310 ms, avg took 0.031 ms
```

IMHO, no need to hold an instance in this case.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Fixed. Agree. Removed that code. It's "over defensive".

cassandra-mesos-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManagerTest.java
dmitrypekar: Don't want to introduce extra class/interface like JsonFetcher or spawn embedded server in UT. IMHO, we are ok to stay with overriding, as the most simplest way to implement UT in this case.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Fixed.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraScheduler.java
dmitrypekar: IMHO, this is engineering overkill. It will not add anything useful, except additional complexity.
The overhead of evaluating the check "do we need to fetch" is minimal. Let me collect the actual numbers ...

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Reverted. Need to override method in UT.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: Reverted. It is used from Main in from other package.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraScheduler.java
dmitrypekar: Please see the timing of syncSeeds. Where are 2 cases: 1. when fetching is skipped because delay is not passed. 2. when seeds are actually fetched. Timing is following:
```
cassandraCluster.getSeedManager().syncSeeds() took 265 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 25 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 1 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
cassandraCluster.getSeedManager().syncSeeds() took 0 ms
```
Case 1. doesn't take a lot of time in any case. Case 2. could potentially take several seconds to fetch is the endpoint is slow. Even if it will go to timeout, it should not take more than a minute for each dc. Hm.. It could potentially block the scheduler. Need to think ..

dmitrypekar: There is rationale in moving syncing seeds to another thread in case then http connection blocks.
Fixed. Updated SeedManager to use ScheduledExecutorService.

BenWhitehead: Correct, we don't want to the offer cycle to control or be impacted by the need of the seed manager to refresh it's info.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
BenWhitehead: My comment is for Jackson in general, in the case of this codebase, there isn't much classpath scanning or initialization that has to take place but in many other code bases there can be a lot, and trying to do this code that has more load on it can cause a lot of slow down and GC pressure .  It considered best practice to use a single instance of object mapper for each configuration profile.

BenWhitehead:
@dmitrypekar See [node](https://github.com/mesosphere/cassandra-mesos/pull/120#discussion_r34816037) about ObjectMapper.

Once that's fixed, please rebase squash this whole PR in prep for merge.

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/SeedManager.java
dmitrypekar: fixed.

dmitrypekar:
Squashed to a single commit. Could be merged now.

BenWhitehead:
Merged, thanks @dmitrypekar.

BenWhitehead:
I've got a preliminary change to the package/config definition for the universe that I'll and can get a pr of soon so that we can provide this new config for dcos installs.

PR #125, opened 2015-07-28
Set LD_LIBRARY_PATH so Cassandra runs on Test/Continuous Clusters

gabrielhartmann:


cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
```diff
@@ -93,7 +93,8 @@
     );
 
     private static final Map<String, String> executorEnv = unmodifiableHashMap(
-        tuple2("JAVA_OPTS", "-Xms256m -Xmx256m")
+        tuple2("JAVA_OPTS", "-Xms256m -Xmx256m"),
+        tuple2("MESOSPHERE_LD_LIBRARY_PATH", "/opt/mesosphere/lib")
```
elingg: On non dcos clusters (most major linux distributions), LD library path is typically located under "/usr/local/lib". I'm not very familiar with this code, but I wanted to double check that the default value is correct here.

elingg:
One comment which is pretty important. Other than that, LGTM. Any objections @BenWhitehead  before merging?

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
BenWhitehead: This value is appended to the path at executor start time. You're right that it's normally not there and will only be there in DCOS.

I want to test this out on DCOS and mesos to make sure that it works in both locations then I'll merge it.

gabrielhartmann: This has been tested on DCOS and mesos clusters now.

PR #135, opened 2015-08-21
restore_basic2

dmitrypekar:
If recreated branch and PR for restore_basic functionality.
Please review & merge.


elingg:
hello @dmitrypekar, I will do my best to look into this tomorrow.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @NotNull
+    private final String backupDir;
+
+    public BackupManager(@NotNull final JmxConnect jmxConnect, @NotNull final String backupDir) {
+        this.jmxConnect = jmxConnect;
+        this.backupDir = backupDir;
+    }
+
+    public void backup(@NotNull final String keyspace) throws IOException {
```
elingg: Could we have a more descriptive name for the keyspace here, i.e. "cassandra-keyspace-" + System.currentTimeMillis()

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
```diff
@@ -0,0 +1,127 @@
...
+
+        LOGGER.info("Copying backup of keyspace {}", keyspace);
+        copyKeyspaceSnapshot(snapshot, keyspace);
+
+        LOGGER.info("Clearing snapshot of keyspace {}", keyspace);
+        storageServiceProxy.clearSnapshot(snapshot, keyspace);
+    }
+
+    private void copyKeyspaceSnapshot(final String snapshot, final String keyspace) throws IOException {
+        final List<String> tables = jmxConnect.getColumnFamilyNames(keyspace);
```
elingg: Quick question- is this typically how Cassandra backup is done? We iterate through all tables and take the snapshot of each?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
```diff
@@ -0,0 +1,127 @@
...
+        for (final String table : tables)
+            copyTableSnapshot(snapshot, keyspace, table);
+    }
+
+    @SuppressWarnings("ResultOfMethodCallIgnored")
```
elingg: Why are we suppressing warnings here?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
```diff
@@ -0,0 +1,127 @@
...
+            for (final File file : files) {
+                if (file.isFile()) {
+                    Files.copy(file.toPath(), new File(destDir, file.getName()).toPath(), StandardCopyOption.REPLACE_EXISTING);
+                }
+            }
+        }
+    }
+
+    private File findTableSnapshotDir(final String keyspace, final String table, final String snapshot) {
+        final File dataDir = new File(jmxConnect.getStorageServiceProxy().getAllDataFileLocations()[0]);
```
elingg: We should probably name this method createTableSnapshotDir

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private void copyKeyspaceSnapshot(final String snapshot, final String keyspace) throws IOException {
+        final List<String> tables = jmxConnect.getColumnFamilyNames(keyspace);
+        for (final String table : tables)
+            copyTableSnapshot(snapshot, keyspace, table);
+    }
+
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    private void copyTableSnapshot(final String snapshot, final String keyspace, final String table) throws IOException {
```
elingg: The method arguments should be @NotNull here as well right?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
```diff
@@ -0,0 +1,127 @@
...
+        final File[] files = srcDir.listFiles();
+        if (files != null) {
+            for (final File file : files) {
+                if (file.isFile()) {
+                    Files.copy(file.toPath(), new File(destDir, file.getName()).toPath(), StandardCopyOption.REPLACE_EXISTING);
+                }
+            }
+        }
+    }
+
```
elingg: The method parameters should be @NotNull here as well right?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
```diff
@@ -0,0 +1,127 @@
...
+        final File tableDir = findTableDir(keyspaceDir, table);
+        final File snapshotDir = new File(tableDir, "/snapshots/" + snapshot);
+        if (!snapshotDir.exists()) throw new IllegalStateException("Snapshot dir does not exist: " + snapshotDir);
+
+        return snapshotDir;
+    }
+
+    private File findTableDir(final File keyspaceDir, final String table) {
+        final File[] files = keyspaceDir.listFiles();
```
elingg: The method parameters should be @NotNull here as well right? I am working toward consistency here.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+            LOGGER.info("Restoring backup of {}/{}", keyspace, table);
+            restoreTableSnapshot(keyspace, table);
+
+            LOGGER.info("Reloading SSTables for {}/{}", keyspace, table);
+            jmxConnect.getColumnFamilyStoreProxy(keyspace, table).loadNewSSTables();
+        }
+    }
+
```
elingg: The method parameters should be @NotNull here as well right? Why are we suppressing warnings here?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java
```diff
@@ -0,0 +1,109 @@
...
+        restoreFeature = executorService.submit(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    LOGGER.info("Starting restore on keyspace {}", keyspace);
+                    keyspaceStarted();
+
+                    final BackupManager backupManager = new BackupManager(checkNotNull(jmxConnect), backupDir);
+                    backupManager.restore(keyspace, truncateTables);
+
```
elingg: Let's use constants for SUCCESS and FAILURE since they are used in multiple places

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java
```diff
@@ -0,0 +1,109 @@
...
+                    keyspaceFinished("SUCCESS", keyspace);
+                } catch (final Exception e) {
+                    LOGGER.error("Failed to restore keyspace " + keyspace, e);
```
elingg: If there is a failure, we only want to log this?

elingg:
review in progress @dmitrypekar. Feel free to take a look at comments so far.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: Sure. Fixed.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: Fixed.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: Fixed.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: Fixed.

dmitrypekar: We are suppressing warnings, because we are ignoring the result of mkdirs() here.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: We are ignoring the result of mkdirs() invokation here. Even if it fails we'll get IOException, so no need to check the result directly.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: Yes, that is a typical way. Usually backup is done via nodetool snapshot:
http://docs.datastax.com/en/cassandra/2.0/cassandra/tools/toolsSnapShot.html

After backup is done, user needs to copy snapshot of each keyspace/table pair to
the "reliable" location. This seems to be the most generic scenario of the backup.
Scenario of the restore is opposite.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java
dmitrypekar: Sure. Added AbstractNodeJob.SUCCESS|FAILURE constants. Usages of appropriate string literals replaced with constants.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java
dmitrypekar: We are also reporting failure back to Scheduler using `keyspaceFinished(FAILURE, keyspace)` invocation, so the fact of failure would not be lost. But the reason of failure would be logged only to executor's log. We are not passing the reason back to Scheduler. That is the common way how all NodeJobs were designed. I just followed that approach.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: @elingg I'm sorry, but it seems, that I didn't understand the point.
Do you want to rename keyspace parameter to cassandraKeyspace or something else?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: Snapshot dir should exists on the moment of invocation of this method.
It should be already created by `storageServiceProxy.takeSnapshot(snapshot, keyspace)` call.
So, imho, 'find' is the appropriate name.

dmitrypekar:
Hi @elingg I've answered\resolved all the comments. Please go ahead with further review/merge.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
elingg: Sorry this was not clear, @dmitrypekar, I think the snapshot name should be more descriptive rather than s-<timestamp>

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
elingg: We should handle this exception instead of having it trickle up since it seems fairly serious. We should think about what the appropriate error handling behavior should be.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
elingg: We should handle this exception instead of having it trickle up since it seems fairly serious. We should think about what the appropriate error handling behavior should be. I think we should not throw the IOException in the backup method, but at the very least instead catch it and do the appropriate handling.

cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java
```diff
@@ -0,0 +1,191 @@
...
+                                    invocations.add("takeSnapshot");
+                                    return null;
+                                case "clearSnapshot":
+                                    invocations.add("clearSnapshot");
+                                    return null;
+                                case "truncate":
+                                    invocations.add("truncate");
+                                    return null;
+                                case "getAllDataFileLocations":
+                                    return new String[]{dataDir};
```
elingg: Let's keep this as a single string for now until we actually add the support for multiple locations. It is confusing hard coding index 0 elsewhere.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java
```diff
@@ -72,11 +77,13 @@ public void run() {
                     LOGGER.info("Starting backup on keyspace {}", keyspace);
                     keyspaceStarted();
 
-                    checkNotNull(jmxConnect).getStorageServiceProxy().takeSnapshot(backupName, keyspace);
-                    keyspaceFinished("SUCCESS", keyspace);
+                    final BackupManager backupManager = new BackupManager(checkNotNull(jmxConnect), backupDir);
+                    backupManager.backup(keyspace);
+
```
elingg: why is this marked as final?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java
```diff
@@ -72,11 +77,13 @@ public void run() {
...
+                    keyspaceFinished(SUCCESS, keyspace);
                 } catch (final Exception e) {
                     LOGGER.error("Failed to backup keyspace " + keyspace, e);
-                    keyspaceFinished("FAILURE", keyspace);
```
elingg: Do we really want to continue if there is a failure for a given keyspace or do we want to exit?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java
```diff
@@ -0,0 +1,109 @@
...
+                    final BackupManager backupManager = new BackupManager(checkNotNull(jmxConnect), backupDir);
+                    backupManager.restore(keyspace, truncateTables);
+
+                    keyspaceFinished(SUCCESS, keyspace);
+                } catch (final Exception e) {
+                    LOGGER.error("Failed to restore keyspace " + keyspace, e);
+                    keyspaceFinished(FAILURE, keyspace);
+                } finally {
+                    startNextKeyspace();
+                }
```
elingg: Do we really want to continue if there is an error? Should we instead exit?

cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeoutException;
+
+import static junit.framework.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class BackupManagerTest {
+    public static final String KEYSPACE = "k";
```
elingg: Could we add more tests cases here to ensure everything successfully backs up and restores?

cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+        private StorageServiceMBean newStorageService() {
+            return (StorageServiceMBean) Proxy.newProxyInstance(getClass().getClassLoader(), new Class[]{StorageServiceMBean.class},
+                    new InvocationHandler() {
+                        @Override
+                        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+                            String name = method.getName();
+                            switch (name) {
+                                case "takeSnapshot":
+                                    invocations.add("takeSnapshot");
```
elingg: question- why are we returning null here for each case?

elingg:
hello @dmitrypekar, thanks for the great work and addressing my initial comments! I added more comments. Also, how was this tested? Is the typical procedure for Cassandra that you test both on DCOS and on Mesos?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: No prob. Changed name from s-$timestamp -> snapshot-$timestamp.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: IOException is handled on the level of NodeBackupJob/NodeRestoreJob. If it is catched there, following code is executed:
`LOGGER.error("Failed to restore keyspace " + keyspace, e);
 keyspaceFinished(FAILURE, keyspace);
`
Do you think this is not appropriate and some additional actions should be taken?

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
dmitrypekar: I am not sure, what kind of handling should we do there?
In situations like low-disk space, for example?
Imho, the only thing we can do is to fail the entire job in this case.

cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java
dmitrypekar: Ok.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java
dmitrypekar: As I've understood it is a generic coding style applied here, to use final for each variable, that is not modified in the code.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java
dmitrypekar: I think, both approaches could be applied here. But some keyspaces could be already processed before the failure. And we can't guarantee atomicity here, so the approach "do the best to backup/restore as much as possible", imho could be used here.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java
dmitrypekar: The same thoughts as for NodeBackupJob.

cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java
dmitrypekar: This test covers full basic scenario of backup/restore functions. Do you think we need some additional scenarios or do we need tests covering BackupManager internals like findSnapshotDir, etc ?

cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java
dmitrypekar: This is an appropriate way to handle void return type in Proxy InvocationHandler implementation.
Just the same way as when you invoke void method via Reflection it will return null.

dmitrypekar:
Hi @elingg 
Thank you for your review and comments. I've added answer to some of them and pushed fixes.
The implementation was tested on the Mesos cluster without DCOS. I can test on DCOS also if required, but suppose, that those changes are not DCOS-related so could be tested just on Mesos cluster.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java
elingg: I see, thanks. You are correct. I am trying to be consistent with the style of the existing code.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
elingg: perfect thanks

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
elingg: Hm, I think partial failure of the job is okay so long as we do not continue trying to backup the keyspace and move on to the next one. Is my understanding correct there? Also, if there is an exception I think we should catch it immediately and partially fail the job there instead of suppressing the warning. We should catch the error sooner rather than later if possible.

cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java
elingg: Same as my response above:

```Hm, I think partial failure of the job is okay so long as we do not continue trying to backup the keyspace and move on to the next one. Is my understanding correct there? Also, if there is an exception I think we should catch it immediately and partially fail the job there instead of suppressing the warning. We should catch the error sooner rather than later if possible.```

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java (hunk 13 above)
Comment (elingg) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java]: Okay, I think this approach is reasonable as long as we fail fast (stop processing the keyspace and move onto the next one). Also, is the user notified of this failure? Are then able to try again with the specific keyspace? Just wanted to clarify

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java (hunk 14 above)
Comment (elingg) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java]: same response as above

File: cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java (hunk 15 above)
Comment (elingg) [inline cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java]: Ideally, we would also test the internals to make sure the data is backed up and restored and verify that all data is intact. Also, more than one basic scenario test case would be nice.

**Comment by elingg** [comment]
Thanks @dmitrypekar! Sorry for the delay as I was in scrum training. I left a few more comments to be addressed. Once those are addressed, I can do a final review. Definitely let me know if you have questions or input.

**Comment by elingg** [comment]
The PR for the readme is here, correct? https://github.com/mesosphere/cassandra-mesos/pull/134

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java (hunk 3 above)
Comment (dmitrypekar) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java]: Yes, you are right, partial job failure would move job to the next keyspace.

Let me clarify the situation with suppressing. We are suppressing the fact, that we ignore File.mkdirs() result. Code checkers highlight this as warning, because usually this indicates
an error, but not in our case. The reason of ignoring the result of File.mkdirs() is following:

File.mkdirs() could fail in several cases. Two important for us are:
1. dir already exists. In that case we should continue without error;
2. dir can't be created because of FS error. In that case we should fail;

We are ignoring File.mkdirs() result. So if 1. takes place, we would not fail. And if 2. we would fail with IOException on Files.copy(...) statement. That is the designed behavior.

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java (hunk 8 above)
Comment (dmitrypekar) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java]: Same response as above about warning suppression.

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java (hunk 13 above)
Comment (dmitrypekar) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java]: The user sees the status of each processed keyspace. Also executors.log will contain an error.
But the user will not be able to proceed just from failed keyspace. He would need to start the job from the beginning.

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java (hunk 14 above)
Comment (dmitrypekar) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeRestoreJob.java]: We will continue from next keyspace in case of error.

**Comment by dmitrypekar** [comment]
@elingg Yes, those is PR for readme.
I've answered all comments, except for testcase. Will provide extended BackupManagerTest tomorrow.

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java (hunk 3 above)
Comment (elingg) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java]: Hi Dmitry, I think we should check if the file exists before we try to create it. We should also use Files library as we do in the rest of the code for consistency (that should make it so we don't have to suppress the warnings). see below:

https://github.com/mesosphere/cassandra-mesos/blob/9912b5b8a767b617aca7c6b23d79421d4ba600fc/cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/ProdObjectFactory.java#L95

I also think that we should do the check to make sure the file was created and then throw the exception as needed along with a logging message in the place where it failed to be consistent with the rest of the code: 

https://github.com/mesosphere/cassandra-mesos/blob/9912b5b8a767b617aca7c6b23d79421d4ba600fc/cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/ProdObjectFactory.java#L80

**Comment by elingg** [comment]
One concern remaining is the way the directory is created and being consistent with other Cassandra code. I left a comment above on that, that needs to be addressed. Other than that and the tests, I think we are good to go!

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java (hunk 3 above)
Comment (dmitrypekar) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java]: Used Files.createdDirectories (instead of File.mkdirs()), which implements the above behavior (checks for existence, raise IOException if can't create dir).

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java (hunk 8 above)
Comment (dmitrypekar) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManager.java]: Used Files.createdDirectories (instead of File.mkdirs()), which implements the above behavior (checks for existence, raise IOException if can't create dir).

File: cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java (hunk 15 above)
Comment (dmitrypekar) [inline cassandra-mesos-executor/src/test/java/io/mesosphere/mesos/frameworks/cassandra/executor/BackupManagerTest.java]: Modified backup & restore test cases to be more complex. Added test cases for BackupManager internals.

**Comment by dmitrypekar** [comment]
Hi @elingg 
I've fixed those issue with warning suppression. Also added more test cases to BackupManagerTest.
Please review and merge if appropriate.

File: cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java (hunk 13 above)
Comment (elingg) [inline cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/jmx/NodeBackupJob.java]: I see, let's make sure this is documented.

**Comment by elingg** [comment]
Thanks @dmitrypekar! It's too bad @BenWhitehead is not available for a second pass since I am still learning the code base. 

That said, it looks great to me. Was the latest version tested on Mesos? If yes, I will go ahead and merge.

File: cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto [hunk 17]
```diff
@@ -742,9 +748,13 @@ message NodeJobTask {
...
     /**
-     * Backup name
+     * Backup directory
+     */
+    optional string backupDir = 3;
+    /**
+     * Determine if this is the first node participating in job
      */
-    optional string backupName = 2;
+    optional bool firstNodeInJob = 4;
```
BenWhitehead: What significance does this field have? Why does it need to be tracked?

cassandra-mesos-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraClusterTest.java
```diff
@@ -87,7 +87,7 @@ public void removeExecutor_cleansAllTasksAndExecutorInfo() throws Exception {
         final PersistedCassandraClusterHealthCheckHistory healthCheckHistory = new PersistedCassandraClusterHealthCheckHistory(state);
         final PersistedCassandraClusterJobs jobsState = new PersistedCassandraClusterJobs(state);
         final PersistedCassandraFrameworkConfiguration configuration = new PersistedCassandraFrameworkConfiguration(
-            state, "cassandra.unit-test", 15, 15, "2.1.4", 1.0, 64, 64, 32, 1, 1, "*", ".", true, false,
+            state, "cassandra.unit-test", 15, 15, "2.1.4", 1.0, 64, 64, 32, 1, 1, "*", ".backup", ".", true, false,
```
BenWhitehead: If this is adding a value for the backup directory please change the string to be `./backup` instead of `.backup`.

BenWhitehead:
I would very much like to see a test plan and some results to verify that this works and how it can be ensured to work into the future.  This is going to be a critical set of code that will need lots of tests to ensure it works and continues to work.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
dmitrypekar: That's required because:
- we need to do table truncate during restore;
- table truncate should be done only once;
Please see: 
- io.mesosphere.mesos.frameworks.cassandra.executor.BackupManager#restore 
- io.mesosphere.mesos.frameworks.cassandra.CassandraFrameworkProtos.NodeJobTask#getFirstNodeInJob

cassandra-mesos-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraClusterTest.java
dmitrypekar: Fixed.

dmitrypekar:
Hi Ben,

I've used primarily following test to verify the implementation of backup&restore is working:
# Test plan:
1. Run cassandra mesos with 2 nodes;
2. Create table t.test with some data:
```
CREATE KEYSPACE t WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };
USE t;
CREATE TABLE test (id int PRIMARY KEY, s varchar);

insert into test (id, s) values (1, '1');
insert into test (id, s) values (2, '2');
insert into test (id, s) values (3, '3');
insert into test (id, s) values (4, '4');
insert into test (id, s) values (5, '5');
insert into test (id, s) values (6, '6');
insert into test (id, s) values (7, '7');
insert into test (id, s) values (8, '8');
insert into test (id, s) values (9, '9');
insert into test (id, s) values (10, '10');
```
3. Create backup 'backup';
4. Delete some data:
```
delete from t.test where id in (2,5,7,8);
```
5. Restore backup 'backup';
6. Make sure that deleted data is restored;

dmitrypekar:
Hi @elingg 
Yes, I was testing on the latest mesos available from repo. For now it is 0.23.0.

elingg:
@dmitrypekar, glad to merge! Did you ever test with more than one keyspace?

elingg:
@mattpennathe3rd will be doing a final documentation review this afternoon

dmitrypekar:
@elingg Sure, I've tested with several keyspaces and several tables.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
BenWhitehead: I'm not crazy about the naming here, if it's being used to indicate that the restore first needs to run a truncate ideally that would be the first job ran with a different type.

Something like this:
1. Run `NodeJobTask { jobType = TRUNCATE }`
2. foreach node in nodes `NodeJobTask { jobType = RESTORE, restoreFile = "/local/path/to/backup.bak" }`

Let's keep the smarts about what has to happen in the scheduler which can coordinate the tasks, rather than having tasks figure out some special things based on the global state or other stuff.

BenWhitehead:
have you tested with a large keyspace?

What will happen if there isn't enough disk space to fit the backup file on the machine to be able to restore it?

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
dmitrypekar: @BenWhitehead , 

personally I think, that adding additional type of job is overdesign here.
It will increase the complexity of the implementation without adding
any additional benefit.

Current design includes:
- 2 types of jobs: backup, restore;
- 2 props: backupDir, firstNodeInJob;

Your proposal includes:
- 3 types of jobs: backup, restore, truncate;
- 2 props: backupDir, restoreDir;

Adding a new type of job requires additional control logic
added to 3+ places in code, testing, new review, etc. 

Please take into account, that it will potentially increase time to deliver 
this function in ~3 days (1d for making change, 1d for review and 0.5d for making
review changes), without adding any functional benefit. 

The decision is up to you. Please write, if you prefer design with 3
jobs types and I will implement it. Or may be it would be more practical to
stay with the current design.

dmitrypekar:
@BenWhitehead 

I've tested locally on a table with 1 million of rows. Though, it could not be considered as large dataset.

If no free space would happen, copy of files should fail and the restore for a particular keyspace
would fail. But I didn't tested that scenario directly.

In order to optimize delivery time, my proposal is to split high-load testing in a separate phase,
and merge the changes now. I would run those tests in parallel and provide you with results till the mid of next week (I need to prepare larger cluster for that kind of testing).

If there will be some bugs found (low probability) they could be merged as a separate PR (it would be quicker to do so).

What do you think? Could we merge now and do those testing in a few next days later?

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
elingg: On this one, I agree with @BenWhitehead. We should run truncate as a separate job type in order to have the code broken down into clearer modules. It is harder to read when jobs combined. Also, breaking down code allows for easier error handling and the ability to see where the error has occurred.

elingg:
Hi @dmitrypekar, your testing sounds sufficient for now although I think this needs to be improved in the long term to actually test out the error cases.

We need two more things before merge 
-to address any final docs comments from @mattpennathe3rd 
-to address our comments about breaking out truncate into a separate job

Thanks for the great work and let's get this merged soon!

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
dmitrypekar: @BenWhitehead @elingg 
It seems that I do not understand clearly the scope of change and need to clarify. Do you want me to:
1. add completely separate ClusterTruncateController with API like /cluster/truncate/start|stop|status|abort;
2. integrate new Truncate job in a new RestoreClusterJobHandler adding all those complex coordination logic to invoke Truncate first and then Restore for each node;
Please answer in order for me to be able to continue the implementation.

elingg: Hi @dmitrypekar, you should 2) integrate new Truncate job in a new RestoreClusterJobHandler adding all those complex coordination logic to invoke Truncate first and then Restore for each node

mattpennathe3rd:
@elingg @BenWhitehead @keithchambers @sascala @dmitrypekar 

I made some extensive changes to the documentation, so I strongly encourage thorough review to check for accuracy. I did my best to warn the user about data loss issues after speaking with Ben.

You can find it in a gist here: https://gist.github.com/mattpennathe3rd/bfe8a359950f96d09d18

Let me know what I can do at this point.

cassandra-mesos-model/src/main/proto/io/mesosphere/mesos/frameworks/cassandra/model.proto
dmitrypekar: Hi @elingg,
I've extracted TRUNCATE job and updated logic of NodeTaskClusterJobHandler instead of adding RestoreClusterJobHandler (99% duplicated code). 

Now the Scheduler decides what jobs run where, as you and Ben wanted.
I am going to update the docs now. Please merge after that.

dmitrypekar:
@elingg @BenWhitehead @mattpennathe3rd 
I've integrated the above doc changes.

All required points are addressed now.

Please merge.

elingg:
Ok, sounds good. Merging now, but I would suggest that it will be useful to refactor this piece since it is will be a cleaner implementation to also store the job type.

dmitrypekar:
@elingg Thank you!

PR #136, opened 2015-08-27
attribute2config

dmitrypekar:
Hi,

I've added support of CASSANDRA_RACK, CASSANDRA_DC slave attributes. If those attributes are present in Offer, Scheduler uses their value as RACK/DC properties of CassandraNode.

Please review & merge.

### DIFF
--- a/cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/CassandraFrameworkProtosUtils.java
+++ b/cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/CassandraFrameworkProtosUtils.java
@@ -188,7 +188,8 @@ public final class CassandraFrameworkProtosUtils {
         return ResourceToPortSet.INSTANCE;
     }
 
-    public static Predicate<Resource> containsPort(@NotNull long port) {
+    @NotNull
+    public static Predicate<Resource> containsPort(long port) {
         return new ContainsPort(port);
     }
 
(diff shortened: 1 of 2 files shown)


elingg:
Thanks @dmitrypekar, this seems like a good first step. Is the next step to select offers based on their attributes?

cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java
```diff
@@ -448,6 +451,18 @@ private CassandraNode buildCassandraNode(@NotNull final Protos.Offer offer, fina
         }
     }
 
+    @NotNull
+    private RackDc getRackDc(@NotNull final Protos.Offer offer) {
+        final String rack = attributeValue(offer, RACK_ATTRIBUTE);
```
elingg: Could you add a descriptive comment here? Also, the implementation is unchanged if the agent node attributes are not set, correct?

dmitrypekar: Sure. Added comment with description.

Yes, the behavior is exactly the same if those attributes are not defined in offer (slave node).

cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/CassandraFrameworkProtosUtils.java
```diff
@@ -188,7 +188,8 @@ public static CassandraNodeTask getTaskForNode(@NotNull final CassandraNode cass
         return ResourceToPortSet.INSTANCE;
     }
 
-    public static Predicate<Resource> containsPort(@NotNull long port) {
+    @NotNull
+    public static Predicate<Resource> containsPort(long port) {
```
elingg: small nit, this parameter should be final.

elingg:
One nit above that should be addressed in a subsequent PR, but I'm merging.

elingg:
:+1:

Issue #138, opened 2015-09-06
Invalid cluster name specified in the cassandra.yaml


vimvim:
Executor adds "cassandra." prefix to the cluster name specified in the marathon.json

---------------------------------------------------------------------------------------------------------------------------
marathon.json:
  "env": {
   ....
    "CASSANDRA_CLUSTER_NAME": "ClusterName1",

---------------------------------------------------------------------------------------------------------------------------
cassandra.yaml ( on the worker node ):
....
cluster_name: cassandra.ClusterName1
...
---------------------------------------------------------------------------------------------------------------------------

elingg:
I believe this is as expected. Is it causing any issues?

vimvim:
Yes, this problem is cause serious configuration issue because we unable to setup replication with the external Cassandra cluster ( native Cassandra - without cassandra-mesos ). External cluster has name "ClusterName1" ( not real name - only for example). This is production cluster,so we unable to change this name to another. At the same time we can't specify this cluster name for new cassandra-mesos cluster ( cassandra-mesos adds "cassandra." prefix )

kensipe:
we had an engineering discussion on this today... there is framework work needed in order to provide this ability and I've added a task internally to be scheduled.  We are open to a PR if you would like to accelerate that.   If you do work on this, it is important to provide previous naming for clusters from the previous version in addition to adding new behavior to new clusters.

elingg:
@BenWhitehead created a PR for this, thanks! With the latest master code it will work, @vimvim

PR #142, opened 2015-09-25
Decouple framework name and cluster name in cassandra.yaml

BenWhitehead:


BenWhitehead:
This PR should make it possible to do what is mentioned in #138

BenWhitehead:
@elingg please review

elingg:
Thanks @BenWhitehead. I will take a look ASAP

cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
```diff
@@ -134,7 +136,8 @@ private static int _main() throws UnknownHostException {
         final long      healthCheckIntervalSec      = Long.parseLong(       Env.option("CASSANDRA_HEALTH_CHECK_INTERVAL_SECONDS").or("60"));
         final long      bootstrapGraceTimeSec       = Long.parseLong(       Env.option("CASSANDRA_BOOTSTRAP_GRACE_TIME_SECONDS").or("120"));
         final String    cassandraVersion            =                       "2.1.4";
-        final String    frameworkName               = frameworkName(        Env.option("CASSANDRA_CLUSTER_NAME"));
```
elingg: Style needs to be consistent here,   for example mesosUser is on one line ```  final String    mesosUser                   =                       Env.option("MESOS_USER").or("");```

cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
```diff
@@ -134,7 +136,8 @@ private static int _main() throws UnknownHostException {
...
+        final String    clusterName                 =                       clusterNameOpt.or("cassandra");
+        final String    frameworkName               = frameworkName(clusterNameOpt);
         final String    zkUrl                       =                       Env.option("CASSANDRA_ZK").or("zk://localhost:2181/cassandra-mesos");
```
elingg: fix indentation for frameworkName

cassandra-mesos-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/PersistedCassandraFrameworkConfigurationTest.java
```diff
@@ -49,4 +61,81 @@ public void testGetDefaultRackDc() {
...
+            "./backup",
+            ".",
+            true,
+            true,
+            "rack1",
+            "dc1",
+            Collections.<ExternalDc>emptyList(),
+            "clusterName"
+        );
+
```
elingg: Question here: I'm unclear on how this test case works. Based on the code I see above, if you set the clusterName in the environment variables, then the framework name should be cassandra.clusterName. This seems to be a separate case of using a persisted configuration. Could you elaborate?

elingg:
Thanks for the PR @BenWhitehead. Looks great, a few small comments and a bigger question regarding configuration

cassandra-mesos-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/PersistedCassandraFrameworkConfigurationTest.java
BenWhitehead: This code has nothing to do with the environment variables in `Main`.  This test is specifically there to test and ensure that when initialized that `clusterName` and `frameworkName` can now be two different values since that is the desire from the user.

The test case above this one, reads an existing Persisted Configuration and makes sure that the cluster name stays the same value as was computed before while the proto is "upgraded".

cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
BenWhitehead: The option for cluster is an existing parameter, however it used to be overloaded to configure framework name (by prepending `cassandra.`) and then that framework name would be used as the cluster name.

This change makes it so that `CASSANDRA_CLUSTER_NAME` can be provided to specify the cluster name (any string the user wants, if nothing is specified a default value of `cassandra` is used), the framework name will then prepend `cassandra.` to be consistent with the existing framework naming scheme.

cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
BenWhitehead: fixed

cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
elingg: I understand the change. I was actually referring to consistent coding style clusterNameOpt should not be a separate variable for consistency. It is a nit, but I would prefer to be as clean as possible.

cassandra-mesos-scheduler/src/test/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/PersistedCassandraFrameworkConfigurationTest.java
elingg: It makes sense what you are doing in these two tests, but I have a separate source of confusion.

Based  on the logic in main, it seems that if the user configures the clusterName environment variable then the framework name becomes cassandra.<cluster-name>, but it seems that here you are not using environment variables for configuration and updating the configuration by using the persisted configuration. Does the user have two options- PersistedConfiguration or environment variables?
        if (clusterName.isPresent()) {
            return "cassandra." + clusterName.get();
        } else {
            return "cassandra";
        }

BenWhitehead: The environment variables are only used to initialize the persisted config object the first time the framework starts.  After that, the only environment variables that are used are the zk url for the state store and the zk url to locate the leading master.  The persisted config is then used for everything else.

cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
BenWhitehead: The nit would introduce and additional allocation of the variable, all the other variables are only initialized once because they are only read once.

elingg: Discussed with @BenWhitehead

elingg:
Discussed with @BenWhitehead and this was tested on a clsuter. Looks good!

PR #146, opened 2015-10-01
Dynamic reservations

edgefox:


keithchambers:
@edgefox thank you for this! Our team at Mesosphere will review next week.

cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/ProtoUtils.java
```diff
@@ -59,17 +60,29 @@ public static TaskID taskId(@NotNull final String name) {
 
     @NotNull
     public static Resource cpu(final double value, @NotNull final String role) {
-        return scalarResource("cpus", value, role);
+        return scalarResource("cpus", value, role, false, null);
+    }
+
+    public static Resource reserveCpu(final double value, @NotNull final String role, @NotNull final String principal) {
```
BenWhitehead: Add `@NotNull` to the return of this method

cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/ProtoUtils.java
```diff
@@ -59,17 +60,29 @@ public static TaskID taskId(@NotNull final String name) {
...
+        return scalarResource("cpus", value, role, true, principal);
     }
 
     @NotNull
     public static Resource mem(final double value, @NotNull final String role) {
-        return scalarResource("mem", value, role);
+        return scalarResource("mem", value, role, false, null);
+    }
+
+    public static Resource reserveMem(final double value, @NotNull final String role, @NotNull final String principal) {
```
BenWhitehead: Add `@NotNull` to the return of this method

cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/ProtoUtils.java
```diff
@@ -59,17 +60,29 @@ public static TaskID taskId(@NotNull final String name) {
...
+        return scalarResource("mem", value, role, true, principal);
     }
 
     @NotNull
     public static Resource disk(final double value, @NotNull final String role) {
-        return scalarResource("disk", value, role);
+        return scalarResource("disk", value, role, false, null);
+    }
+
+    public static Resource reserveDisk(final double value, @NotNull final String role, @NotNull final String principal) {
```
BenWhitehead: Add `@NotNull` to the return of this method

cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/ProtoUtils.java
```diff
@@ -88,8 +101,9 @@ public static Resource ports(@NotNull final Iterable<Long> ports, @NotNull final
     }
 
     @NotNull
-    public static Resource scalarResource(@NotNull final String name, final double value, @NotNull final String role) {
-        return Resource.newBuilder()
+    public static Resource scalarResource(@NotNull final String name, final double value,
```
BenWhitehead: If wrapping is needed please place each parameter on it's own line, see https://github.com/mesosphere/cassandra-mesos/blob/c16e97f093750039791526eeb52bd7b97a9c4c74/cassandra-mesos-model/src/main/java/io/mesosphere/mesos/util/ProtoUtils.java#L126-126.

BenWhitehead:
After deeper review I think there are some inherent design issues with the approach implemented in this pull request.

The philosophy of cassandra-mesos so far has been to avoid attempting to do feature detection against mesos due to fragility and maintenance concerns. Attempting to read the version from the master is very very fragile code (endpoints can go away) and should be removed and not included.  Then the project documentation should be updated to declare a new minimum mesos version of 0.24.0 if dynamic reservations are to be used and left at that.

A new parameter has been added to indicate whether resources reservations should be made or not `CASSANDRA_RESERVE`.  This parameter should be used to signal to cassandra-mesos whether it should attempt to create resource reservations.  Since role and principal are required in order to create a reservation the user should be expected to provide values, no default principal should be assumed.

The new method `reserveResources` added to `CassandraScheduler` effectively just attempts to reserve resources on the node as long as the number of nodes with reservations is lower than the target number of nodes without coordinating that the nodes are actually launched on the node with the reservation. `CassandraCluster` is the component with the information necessary to determine if a resource reservation should be made on the node. I would expect to see a new similar to [this one](https://gist.github.com/BenWhitehead/8267c72e4e894c2e2b29) added to `CassandraCluster` that `CassandraScheduler` would then call during `CassandraScheduler#evaluateOffer`.

[`CassandraCluster#_getTasksForOffer`](https://github.com/mesosphere/cassandra-mesos/blob/5c13f1d601d4b82b3f325324d46a40c3986da8d6/cassandra-mesos-scheduler/src/main/java/io/mesosphere/mesos/frameworks/cassandra/scheduler/CassandraCluster.java#L1116) should then be updated to only launch executors on nodes that have resources reservations.

There is no code that Unreserves resources when a node is terminated, it's not okay to leave a reservation in place when a node is moved to a new host.

No unit or integration tests were added to verify that the new code actually does what it's supposed to and results in the correct state changes.

zhitaoli:
Hi, any update on this PR? We are looking into running this framework w/ added support of persistent volumes, and I think running on reserved resources is a prerequisite?

triclambert:

PR #149, opened 2015-10-23
Fix NPE when startup fails and 'process' is null

nickbp:
This failure case results in emitting a TASK_ERROR status update at L155, so it looks like we should skip the TASK_RUNNING message entirely in this case? Alternate option could be to send the message, but with a different or unset value for the PID.

### DIFF
--- a/cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/CassandraExecutor.java
+++ b/cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/CassandraExecutor.java
@@ -160,7 +160,9 @@ public final class CassandraExecutor implements Executor {
                         serverTask = task;
                         jmxConnect = objectFactory.newJmxConnect(cassandraServerRunTask.getJmx());
                         startCheckingHealth(driver, jmxConnect, cassandraServerRunTask.getHealthCheckIntervalSeconds());
-                        if (process != null) {
+                        if (process == null) { 
+                            ExecutorUtils.serverProcessNotRunning(driver, task);
+                        } else {
                             driver.sendStatusUpdate(taskStatus(serverTask, TaskState.TASK_RUNNING,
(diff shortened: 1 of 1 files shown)


cassandra-mesos-executor/src/main/java/io/mesosphere/mesos/frameworks/cassandra/executor/CassandraExecutor.java
```diff
@@ -160,12 +160,14 @@ public void launchTask(final ExecutorDriver driver, final TaskInfo task) {
                         serverTask = task;
                         jmxConnect = objectFactory.newJmxConnect(cassandraServerRunTask.getJmx());
                         startCheckingHealth(driver, jmxConnect, cassandraServerRunTask.getHealthCheckIntervalSeconds());
-                        driver.sendStatusUpdate(taskStatus(serverTask, TaskState.TASK_RUNNING,
-                            SlaveStatusDetails.newBuilder()
-                                .setStatusDetailsType(SlaveStatusDetails.StatusDetailsType.CASSANDRA_SERVER_RUN)
-                                .setCassandraServerRunMetadata(CassandraServerRunMetadata.newBuilder()
-                                    .setPid(process.getPid()))
-                                .build()));
```
elingg: Hi @nickbp, shouldn't we follow the standard in the code base and exit with an error as we do elsewhere?

```  if (process == null) {
            ExecutorUtils.serverProcessNotRunning(driver, task);
            return;
        }```

nickbp: Makes sense. I had it just call serverProcessNotRunning() without the return call, so that a log call at the end of the function is still reached.

elingg: Cool, sounds good to me

triclambert:

PR #153, opened 2015-11-04
Speeding up the cluster start up

olegkovalenko:
Hey,

Motivation:

Current cassandra-mesos master branch delays next server task start
by predefined interval (for instance 2 minutes through
CASSANDRA_BOOTSTRAP_GRACE_TIME_SECONDS), this delay could be
minimized.  Spin up cluster faster by allowing to start next server
task only when all other Cassandra nodes joined to cluster and
operate normally.

Modification:

Changes for CassandraScheduler, implemented ableToLaunchServerTask,
to replace call to canLaunchServerTask, that will
return true when all CassandraNode nodes, that have server task and target
state is run, have latest health check that satisfies
CassandraCluster.isNodeHealthyJoinedAndOperatingNormally
otherwise don't allow to start next server task (return false). 

No changes for CassandraExecutor.

CASSANDRA_HEALTH_CHECK_INTERVAL_SECONDS has to be set to 
small value like 15 - 25 seconds in order to get health checks often.

Possible improvements that could be made for CassandraExecutor is
when starting cassandra server spin up additional thread
that will poll Cassandra status through JMX and once its
up and operation normally sendStatusUpdate containing
health check details. After then stop checking. Default
health checking thread will remain active. Difference
between additional and default health checking thread:
- additional thread has small health check interval (like 15 seconds) while
  default thread uses interval specified through env variable CASSANDRA_HEALTH_CHECK_INTERVAL_SECONDS
- additional thread will send health check via sendStatusUpdate, 
  where as default thread will send updates via sendFrameworkMessage
- additional thread will stop polling once node join to cluster and operation normally

Result:
Faster startup time of cluster, almost by half time it used to be (depending on CASSANDRA_BOOTSTRAP_GRACE_TIME_SECONDS and CASSANDRA_HEALTH_CHECK_INTERVAL_SECONDS).

### DIFF
--- a/cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
+++ b/cassandra-mesos-framework/src/main/java/io/mesosphere/mesos/frameworks/cassandra/framework/Main.java
@@ -134,7 +134,6 @@ public final class Main {
         final long      resourceDiskMegabytes       = Long.parseLong(       Env.option("CASSANDRA_RESOURCE_DISK_MB").or("2048"));
         final long      javaHeapMb                  = Long.parseLong(       Env.option("CASSANDRA_RESOURCE_HEAP_MB").or("0"));
         final long      healthCheckIntervalSec      = Long.parseLong(       Env.option("CASSANDRA_HEALTH_CHECK_INTERVAL_SECONDS").or("60"));
-        final long      bootstrapGraceTimeSec       = Long.parseLong(       Env.option("CASSANDRA_BOOTSTRAP_GRACE_TIME_SECONDS").or("120"));
         final String    cassandraVersion            =                       "2.1.4";
         final String    clusterName                 =                       clusterNameOpt.or("cassandra");
         final String    frameworkName               =                       frameworkName(clusterNameOpt);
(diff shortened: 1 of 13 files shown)


mohitsoni:
@olegkovalenko Thanks for the PR!

Does the proposed implementation prevent the framework from launching new nodes, until all the previously launched nodes passes health check and joins the cluster ? If yes, can one bad node prevent framework from spinning up the cluster forever ?

Apart from that, the proposed PR should help reduce the cluster startup time, by eliminating the static time delays.

olegkovalenko:
> Does the proposed implementation prevent the framework from launching
> new nodes, until all the previously launched nodes passes health
> check and joins the cluster ? 

Yes

> If yes, can one bad node prevent framework from spinning up the cluster forever ?

For me its hard to predict all possible cases of bad node behavior.
In general depends on lifetime of a bad node, if it eventually dies then 
nothing prevents from starting next node once bad node dies, otherwise 
your point is valid. Perhaps manual intervention is required from the operator 
to handle bad node. 

On the other hand it seems to me that current master branch doesn't
handle bad node, it just spins up next node once it has enough
resources and 2 minutes elapsed since previous launch, so not sure
that next node successfully joins the cluster.

triclambert:

PR #161, opened 2015-12-19
Make cassandra version configurable via build properties

mcoffin:
This removes the hard-coded cassandra version, and allows it to be picked up from the `cassandra-mesos.properties` file generated by `properties-maven-plugin`.

Implemented as a blocker before mcoffin#1.


triclambert:

